PR #13, opened 2021-08-25
feat(store-sdk): add cart

tlgimenes:
Adds cart to base.store. More info at: https://github.com/vtex/faststore/pull/913


igorbrasileiro:
Can you create other PR just to add the cart provider? Now we can't measure the impact of this feature.

igorbrasileiro (changes requested):
Can you decouple the cart feature from the other stuff, doing this we can measure the impact of the `cart provider`.

tlgimenes:
> Can you decouple the cart feature from the other stuff, doing this we can measure the impact of the `cart provider`.

all this code only changes cart related code

igorbrasileiro:
I realized that the product image inside the minicart is loading more than once.

igorbrasileiro:
Maybe I've found a bug, I added a product and went to the checkout page, but there appear no items, this is intended to?

tlgimenes:
> Maybe I've found a bug, I added a product and went to the checkout page, but there appear no items, this is intended to?

Yes, this is happening because the checkout is setting the cookie on the wrong domain. In production this would should work just fine

igorbrasileiro (commented):
> > Maybe I've found a bug, I added a product and went to the checkout page, but there appear no items, this is intended to?
> 
> Yes, this is happening because the checkout is setting the cookie on the wrong domain. In production this would should work just fine

Ok, looks good.

PR #19, opened 2021-08-31
feat: change base account

tlgimenes:
This PR changes the base account to `bravtexfashionstore`. 

This account should be a showcase of VTEX features and possibilities. They are still adding things to the catalog etc. Once they are done we can merge this


tlgimenes:
Closing untill they upgrade some things on the store

PR #33, opened 2021-09-14
feat: Add search filters

tlgimenes:
This PR adds a very simple faceted filter component so we can filter the page and add tests for page filtering

## How to test it?
Take a look at the tests written. Also, try to toggle the facets yourself at the deploy preview

### DIFF
--- a/cypress/integration/plp.test.js
+++ b/cypress/integration/plp.test.js
@@ -15,13 +15,34 @@ describe('Search page Filters and Sorting options', () => {
     cy.visit(pages.collection, options)
     cy.waitForHydration()
 
-    cy.getById('facet-filter-checkbox').should('exist')
+    cy.getById('facet-filter-header')
+      .first()
+      .click()
+      .getById('facet-filter-checkbox')
+      .should('exist')
   })
 
   it('Applies filters after click', () => {
     cy.visit(pages.collection, options)
     cy.waitForHydration()
 
-    cy.getById('facet-filter-checkbox').should('exist').first().click()
+    cy.getById('facet-filter-header')
+      .first()
+      .click()
+      .getById('facet-filter-checkbox')
+      .first()
+      .click()
+      .then(($checkbox) => {
+        const quantity = $checkbox.attr('data-quantity')
+        const value = $checkbox.attr('data-value')
+
+        // Check if filter applied actually endded up on the url
+        cy.location('pathname').should((loc) => {
+          expect(loc).to.include(`${pages.collection}/${value}`)
+        })
+
+        // Check if filter applied actually brought the number of products it said it would
+        cy.getById('product-link').should('have.length', Number(quantity))
+      })
   })
 })
--- a/cypress/integration/plp.test.js
+++ b/cypress/integration/plp.test.js
@@ -1,0 +1,27 @@
+// eslint-disable-next-line spaced-comment
+/// <reference types="cypress" />
+/**
+ * Cypress tests for testing the Cart module
+ */
+
+import { pages, options } from '../global'
+
+describe('Search page Filters and Sorting options', () => {
+  beforeEach(() => {
+    cy.clearIDB()
+  })
+
+  it('Has filters', () => {
+    cy.visit(pages.collection, options)
+    cy.waitForHydration()
+
+    cy.getById('facet-filter-checkbox').should('exist')
+  })
+
+  it('Applies filters after click', () => {
+    cy.visit(pages.collection, options)
+    cy.waitForHydration()
+
+    cy.getById('facet-filter-checkbox').should('exist').first().click()
+  })
+})
--- a/cypress/integration/plp.test.js
+++ b/cypress/integration/plp.test.js
@@ -1,7 +1,7 @@
 // eslint-disable-next-line spaced-comment
 /// <reference types="cypress" />
 /**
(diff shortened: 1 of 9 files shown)


cypress/integration/plp.test.js
```diff
@@ -0,0 +1,48 @@
+// eslint-disable-next-line spaced-comment
+/// <reference types="cypress" />
+/**
+ * Cypress tests for testing the Cart module
```
victorhmp, line 4:
```suggestion
 * Cypress tests for PLP
```

cypress/integration/plp.test.js
```diff
@@ -0,0 +1,48 @@
...
+      .then(($checkbox) => {
+        const quantity = $checkbox.attr('data-quantity')
+        const value = $checkbox.attr('data-value')
+
+        // Check if filter applied actually endded up on the url
+        cy.location('pathname').should((loc) => {
+          expect(loc).to.include(`${pages.collection}/${value}`)
+        })
+
+        // Check if filter applied actually brought the number of products it said it would
```
victorhmp, line 44:
```suggestion
        // Check if the filter applied actually brought the number of products it said it would
```

src/components/search/FacetedFilter/FacetedFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <div className={styles.container}>
+      {facets
+        .filter((facet) => facet.type === 'TEXT')
+        .map(({ name, values }, it) => (
+          <div key={`${name}-${it}`}>
+            <button
+              className={styles.button}
+              onClick={() => setIndex(it)}
+              data-testid="facet-filter-header"
```
victorhmp, line 25: Are these testIds really necessary? Does Cypress provide selectors such as `getByText` or something like that? I think using these selectors enables us to test things based on what the user sees

src/components/search/FacetedFilter/FacetedFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+            >
+              {name}
+            </button>
+            <ul style={{ display: index !== it ? 'none' : 'block' }}>
```
victorhmp, line 29: Instead of hiding the elements using `display: 'none'`, how about preventing them from rendering? I don't know if it's a good idea to encourage people to hide stuff in the DOM instead of just not putting them there in the first place.

victorhmp (changes requested):

src/components/search/FacetedFilter/FacetedFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+import type { FacetedFilter_FacetsFragment } from '@generated/FacetedFilter_facets.graphql'
+
+import * as styles from './FacetedFilter.module.css'
+
+interface Props {
+  facets: FacetedFilter_FacetsFragment[]
+}
+
+function FacetedFilter({ facets }: Props) {
+  const [index, setIndex] = useState(-1)
```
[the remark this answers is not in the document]
emersonlaurentino: typo: `seSelectedFilter` to `setSelectedFilter`

src/components/search/FacetedFilter/FacetedFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+              {values?.map((item) => {
+                const id = `${name}-${item?.name}`
+
+                return (
+                  <li key={id}>
+                    <input
+                      id={id}
+                      type="checkbox"
+                      checked={!!item?.selected}
+                      onChange={() => toggleFacet(item as any)}
```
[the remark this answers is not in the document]
emersonlaurentino: yep, because the Type Facet is not exported [here](https://github.com/vtex-sites/base.store/blob/e1f716cd3770f4ec70fae3de5aefef7975f322f1/src/sdk/search/Provider.tsx#L52)

src/components/search/FacetedFilter/FacetedFilter.tsx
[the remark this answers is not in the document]
tlgimenes:
```suggestion
  const [selectedFilter, setSelectedFilter] = useState(-1)
```

src/components/search/FacetedFilter/FacetedFilter.tsx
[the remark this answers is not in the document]
tlgimenes: it is because it's not using our graphql layer yet, but it's using IO's and IO's typings are broken

src/components/search/FacetedFilter/FacetedFilter.tsx
tlgimenes: The problem of using `getByText` is that I would pin my tests based on the values itself rather than the structure of the html. This is problematic because this is a starter, and, for each store we change we would have to change the tests

victorhmp: Yeees! You're right, I forgot about the very important "starter" part 😅

src/components/search/FacetedFilter/FacetedFilter.tsx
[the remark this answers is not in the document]
victorhmp: Awesome!! 😄

victorhmp (approved):

PR #37, opened 2021-09-21
feat: Sticky search pagination

tlgimenes:
This PR adds sticky search pagination. 

The idea is that our search uses infinite scroll and, every time the user scrolls down, we need to keep the state of the pagination so once the user navigates on the site and goes back to a search page, he's exactly where he's left.

The before/after this feature can be seen below:
<div style="display: flex; justify-content: space-around; width: 100%">

 
</div>


cypress/support/commands.js
```diff
@@ -32,8 +32,12 @@ Cypress.Commands.add('waitForHydration', () => {
   return cy.get(`[data-testid=react-hydrated]`).should('exist')
 })
 
-Cypress.Commands.add('clearIDB', () => {
-  return indexedDB.deleteDatabase('keyval-store')
+Cypress.Commands.add('clearStorage', () => {
+  cy.window()
+    .then((win) => {
+      win.sessionStorage.clear()
+    })
```
victorhmp:
```suggestion
    .then((window) => {
      window.sessionStorage.clear()
    })
```

cypress/integration/plp.test.js
```diff
@@ -75,3 +75,101 @@ describe('Search page Filters and Sorting options', () => {
...
+
+  it('Sticks to last seen page on plp pagination', () => {
+    cy.visit(pages.collection, options)
+    cy.waitForHydration()
+
+    cy.getById('product-link')
+      .should('exist')
+      .should('have.length.gt', 0)
+      .then(($links) => {
+        const before = $links.length // number of products before showMore is clicked
```
victorhmp, line 117:
```suggestion
        // Number of products before showMore is clicked
        const before = $links.length
```

cypress/integration/plp.test.js
```diff
@@ -75,3 +75,101 @@ describe('Search page Filters and Sorting options', () => {
...
+
+        cy.getById('show-more')
+          .should('exist')
+          .click()
+          .then(() => {
+            let skuIdBeforeNavigate // The skuId of the last product on the page
```
victorhmp, line 123:
```suggestion
            // The skuId of the last product on the page
            let skuIdBeforeNavigate 
```

cypress/integration/plp.test.js
```diff
@@ -75,3 +75,101 @@ describe('Search page Filters and Sorting options', () => {
...
+
+            cy.getById('product-link')
+              .should('have.length.gte', before) // number of products after showMore is clicked should be higher
```
victorhmp, line 126:
```suggestion
              // Number of products after showMore is clicked should be higher
              .should('have.length.gte', before) 
```

src/sdk/search/Provider.tsx
```diff
@@ -1,12 +1,12 @@
-/* eslint-disable @typescript-eslint/restrict-plus-operands */
-import React, { createContext, useMemo, useState } from 'react'
+/* eslint-disable react/prop-types */
```
victorhmp, line 1: Why is this necessary? 🤔

src/sdk/search/Provider.tsx
```diff
@@ -71,7 +79,7 @@ export const SearchProvider: FC<Props> = ({
   children,
   pageInfo,
 }) => {
-  const [pages, setPages] = useState([initalState.page])
+  const [pages, setPages] = useState(() => [initalState.page])
```
victorhmp, line 82: Can we avoid having two functions with names `setPage` and `setPages`? They're being used in close proximity and inside the callback for `useMemo` they're being used in the same scope.

How about calling the innermost one (inside the callback) `setCurrentPage`, or something like that? Just to make them easier to differentiate and improve readability.

victorhmp (changes requested):

src/sdk/search/Provider.tsx
tlgimenes: I have no idea. My eslint is complaining about it

src/sdk/search/Provider.tsx
tlgimenes: what about setPage -> pushPage?

src/sdk/search/Provider.tsx
victorhmp: That's weird, it might be something wrong in our ESLint setup then. The docs for this specific rule is this one: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md.

Since we're using TypeScript, and typing the `SearchProvider` props, this error should not be happening

src/sdk/search/Provider.tsx
victorhmp: Yep, sounds good 😄

src/sdk/search/Provider.tsx
tlgimenes: nice. I'll remove it then

victorhmp: I guess this could stay here! I would just add a comment explaining why we need it

victorhmp (approved):

PR #47, opened 2021-09-24
Add axe tests for a11y 

igorbrasileiro:
This PR adds a11y basic cypress test for the home, collection, and product page.
These tests use cypress-axe with axe-core.

Why the lighthouse isn't sufficient? https://www.prometsource.com/blog/automated-web-accessibility-tools

## How it works? 
Run cypress tests

## How to test it?

## References
https://github.com/component-driven/cypress-axe
https://www.prometsource.com/blog/automated-web-accessibility-tools


src/views/collection/Seo/index.tsx
```diff
@@ -31,6 +32,8 @@ function Seo({
     canonical: `/${slug}/`,
   })
 
+  useAddHead({ title: collectionSeo.title })
```
lariciamota, line 35: Why is it collectionSeo.title and not metadata.title?

igorbrasileiro: You're right, i was looking the useMetadata code, and it return ```tsx { ..., title, ...}```

lariciamota (approved):

cypress/integration/a11y.test.js
```diff
@@ -0,0 +1,29 @@
+describe('Accessibility tests', () => {
+  beforeEach(() => {
+    cy.intercept('/graphql/?operationName=BrowserProductQuery*').as(
+      'productQuery'
+    )
+
+    cy.intercept('/graphql/?operationName=GalleryQuery*').as('galleryQuery')
+  })
+  it('Test Home', () => {
```
emersonlaurentino, line 9: You don't need to say it's a test, it's implicit. I think it's better to just leave `Home`. The same in the next tests.

igorbrasileiro: You're right. Thank you for the comment

emersonlaurentino (approved):

src/views/collection/Seo/index.tsx
tlgimenes: ooops, I'm a little bit late to the party. We use `collctionSeo.title` because. for each collection, the user registers an specific title for that page on our admin

PR #53, opened 2021-09-27
feat: send 'add_to_cart' analytics event

heitorado:
This PR uses the methods defined on the store sdk related to sending/receiving analytics events, to send the 'add_to_cart' event. This event is triggered every time a product is added to the shopping cart, and then received by an analytics provider that stores the data. The data that is sent is a merge between the fields required by SAE analytics and by GA4, so it can be used by both.

## How it works? 
The `ProductDetails.ts` and `ProductSummary.ts` passes the required analytics + cart attributes to the `useBuyButton` hook. When the button is pressed, the attributes are sent as an analytics event, which then is listened by the `AnalyticsHandler` and inserted in the `window.dataLayer` array to be consumed by the Analytics service (GA4 or SAE).
After sending the event with the cart attributes plus some extra attributes required by the analytics, the object needs to be destructured and built again so it only has the fields accepted to make the 'add to cart' mutation on the `store-api` side. This is why a new object is build before sending the product to the `addItem` method.

## How to test it?
`yarn test`. New tests added on `cart.test.js`, covering:
- The analytics event is added when clicking on the buy button at the PLP;
- The analytics event is added when clicking on the buy button at the PDP;
- The `add_to_cart` analytics event is sent with at least the minimum required attributes. 
To test it manually, add something into the shopping cart. Then, open the console and check if the `window.dataLayer` array has the correct analytics data for the added item.

## References
https://www.notion.so/vtexhandbook/vtex-addToCart-8125848d75cb4a34adbfa57199fc4b66
https://www.dumky.net/posts/analytics-and-tag-testing-with-cypress/
https://docs.cypress.io/guides/references/assertions#TDD-Assertions
https://docs.cypress.io/guides/references/assertions#BDD-Assertions
https://www.chaijs.com/api/bdd/#method_property

A big thanks to @icazevedo for the help with GA4 and the great documentation of the issues/tasks; also to @tlgimenes for all the help to understand the codebase.


src/sdk/analytics/index.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { useEffect } from 'react'
+import { useAnalyticsEvent } from '@vtex/store-sdk'
+import type { FC } from 'react'
+
+export const AnalyticsProvider: FC = ({ children }) => {
```
icazevedo, line 5: I think it'd be better if we renamed this component to something like `AnalyticsHandler`. It doesn't provide a context, which is when we usually name things as providers in React. I think leaving it this way may confuse people :)

src/sdk/cart/useBuyButton.ts
```diff
@@ -1,12 +1,34 @@
...
 
-export const useBuyButton = (item: Omit<CartItem, 'id'> | null) => {
+type AdditionalItemProperties = {
+  product_reference_id: string | null
+  sku_reference_id: string | null
+  sku_name: string | null
+}
+
+interface VTEXAddToCartEvent extends AddToCartEvent {
+  data: AddToCartEvent['data'] & {
```
icazevedo, line 16: I figured out that on store-sdk we export an interface called `AddToCartData` that can replace this call. I think it'd be cleaner :)

```suggestion
  data: AddToCartData & {
```

src/sdk/cart/useBuyButton.ts
```diff
@@ -16,10 +38,37 @@ export const useBuyButton = (item: Omit<CartItem, 'id'> | null) => {
         return
       }
 
+      sendAnalyticsEvent<VTEXAddToCartEvent>({
+        type: 'add_to_cart',
+        data: {
+          currency: code,
+          value: item.price * item.quantity,
```
icazevedo, line 45: I think it's worth leaving a `TODO` comment here saying that in the future we may explore more robust ways of calculating the value. Maybe there is some gift item or some discount depending on the quantity. I don't think we should worry about that now, but it's valid to document the possibility.

icazevedo (changes requested):
Great work 👏

src/sdk/analytics/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+  useEffect(() => {
+    window.dataLayer = window.dataLayer || []
+  })
```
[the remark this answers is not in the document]
heitorado: Thanks for the heads up! Still exploring out the world of react hooks 👍🏽

src/sdk/analytics/index.tsx
heitorado: That makes a lot of sense. Done at https://github.com/vtex-sites/base.store/pull/53/commits/d1f33648e138fbb01faeed277229284a22e3405a

src/sdk/analytics/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+  useAnalyticsEvent((event) => {
+    window.dataLayer.push(event)
+  })
+
+  return <> {children} </>
```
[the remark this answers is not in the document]
heitorado: You're right. Removed the `FC` return type and added the `PropsWithChildren<unknown>` type to the argument. Then,   TypeScript's type inference does the trick. Done at https://github.com/vtex-sites/base.store/pull/53/commits/31c27fb1fc2d6a15505afadb3bf1564cdf665d92

src/sdk/cart/useBuyButton.ts
heitorado: Fair enough. Done at https://github.com/vtex-sites/base.store/pull/53/commits/d46f50c9cdb1aeee730a2742744a2f8742c89b98

src/sdk/cart/useBuyButton.ts
heitorado: Great. I believe this solves it? https://github.com/vtex-sites/base.store/pull/53/commits/e84887be8e1813e5866d198221006e7a930f3990

icazevedo (approved):
LGTM, but I miss a deeper test to check if the properties of the event are really there and if they are in the correct format. Is it possible to do it?

gatsby-ssr.js
```diff
@@ -6,23 +6,26 @@ import ErrorBoundary from './src/sdk/error/ErrorBoundary'
...
-        effects={uiEffects}
-      >
-        <SessionProvider>
-          <CartProvider mode="optimistic" onValidateCart={validateCart}>
-            {element}
-          </CartProvider>
-        </SessionProvider>
-      </UIProvider>
-    </TestProvider>
+    <AnalyticsHandler>
```
tlgimenes, line 14: Why not use the standard and call it `AnalyticsProvider` ?

src/components/product/ProductSummary/ProductSummary.tsx
```diff
@@ -77,6 +80,10 @@ export const fragment = graphql`
     slug: linkText
     id: productId
     productName
+    brand
```
tlgimenes, line 83: I think these types have changed since last time. Please rebase.

src/sdk/analytics/index.tsx
```diff
@@ -0,0 +1,17 @@
+import { useEffect } from 'react'
+import { useAnalyticsEvent } from '@vtex/store-sdk'
+import type { PropsWithChildren } from 'react'
+
+export const AnalyticsHandler = ({ children }: PropsWithChildren<unknown>) => {
+  useEffect(() => {
+    window.dataLayer = window.dataLayer || []
```
tlgimenes, line 7: This makes a race condition possible if an analytics event is run before this effect. To elimitate this problem please move the dataLayer creation to the global scope. Also, make sure to make it work durring SSR
```suggestion
if (typeof window !== 'undefined') {
   window.dataLayer = window.dataLayer || []
}

export const AnalyticsHandler = ({ children }: PropsWithChildren<unknown>) => {
  useAnalyticsEvent((event) => {
    window.dataLayer.push(event)
  })
```

tlgimenes (changes requested):

src/sdk/analytics/index.tsx
heitorado: @tlgimenes here is where the `AnalyticsProvider` was renamed to `AnalyticsHandler`

gatsby-ssr.js
heitorado: I agree it would 'match' with the other providers being used, but on a previous comment made by @icazevedo he pointed out that given the component implementation, the name "Provider" would not suit it well. I've tagged you in the conversation thread.

ArthurTriis1 (approved):

emersonlaurentino:
You need make a rebase on your branch

icazevedo (changes requested):

igorbrasileiro (approved):
Great job, @heitorado! 👏 🚀

igorbrasileiro (approved):

ArthurTriis1 (approved):

PR #76, opened 2021-10-08
chore: Adds Envelop plugin system

danzanzini:
This PR adds the [envelop plugin system](https://www.envelop.dev/) to our GraphQL layer. More at: https://vtex-dev.atlassian.net/browse/FSSS-12

## How it works? 
All the <em>enveloping</em> logic is encapsulated at the `server/envelop/index.ts` file. In this file, we may add and configure all plugins we want to add to our GraphQL request. I've already added some **regarding performance and security**.

### Error Handling
As part of this task, a **primitive error handling function was added**. If the error returned from the GraphQL request is a [BadRequestError](https://github.com/vtex/faststore/pull/1034), our server will return 400 instead of the default GraphQL 200. We may improve this solution in the future by supporting other <em>flavors</em> of errors.

## How to test it?
Just enter in the preview URL and see that everything is working fine 😄 

## References
- [Envelop plugin hub](https://www.envelop.dev/plugins)
- [Apollo error handling](https://www.apollographql.com/docs/apollo-server/data/errors/)


igorbrasileiro (approved):
Amazin work @danzanzini !!
Just a tip, what do you think about adding the [`compose-middleware` package](https://www.npmjs.com/package/compose-middleware)?

src/api/graphql.ts
```diff
@@ -40,14 +55,16 @@ const handler = async (
     return
   }
 
+  const enveloped = await getEnveloped(maskError)
```
tlgimenes, line 58: AFAIK this should be run outside the `handler` function, can you confirm it? According to their docs, this is done outside the request cycle (https://www.envelop.dev/docs/getting-started#use-your-envelop)

src/server/envelop/index.ts
```diff
@@ -0,0 +1,33 @@
...
+export const getEnveloped = async (formatError: FormatError) =>
+  envelop({
+    plugins: [
+      useSchema(await getSchema()),
+      useExtendContext(getContextFactory()),
+      enableIf(isProduction, () => useMaskedErrors({ formatError })),
+      useGraphQlJit(),
+      useValidationCache(),
+      useParserCache(),
+      useResponseCache(),
```
tlgimenes, line 31: 👀 This uses memory cache right? If so, we could remove this, right?

src/server/envelop/index.ts
```diff
@@ -0,0 +1,33 @@
...
+const isProduction = ENV === 'production'
+
+export type FormatError = FormatErrorHandler
+
+export const getEnveloped = async (formatError: FormatError) =>
+  envelop({
+    plugins: [
+      useSchema(await getSchema()),
+      useExtendContext(getContextFactory()),
+      enableIf(isProduction, () => useMaskedErrors({ formatError })),
```
tlgimenes, line 27: Isn't it better to always mask the errors and console.log it instead so people don't use error respose on the frontend as logic?

danzanzini: Makes sense. The error's already being logged at dev 👍

danzanzini:
Thanks, @igorbrasileiro. I've added a new Jira task to accomplish this later.

@tlgimenes, I've tackled all your comments. The PR is ready to be reviewed again.

tlgimenes (approved):
LGTM 🎉

PR #81, opened 2021-10-15
Fix: remove duplicate attribute that was causing the product picture to blink when added to cart

heitorado:
Fix the blinking product picture when it was added to cart

## How it works? 
It removes a property which was being updated shortly after, causing a re-rendering of the component.

## How to test it?
Add a product to cart. See that the picture does not blink. You can check it blinks when reverting to the previous commit, or just running in the master branch.


tlgimenes (approved):

igorbrasileiro (approved):
LGTM! But why removing the `id`, the component stopped blinking?

heitorado:
@igorbrasileiro That's because (I believe) that when I sent the `id` in the object, the method who received it was reassigning the `id` attribute. That made the `item` change, which triggered some `useMemo` / `useCallback` / `useEffect` that used the object as a dependency.

PR #98, opened 2021-11-08
feat: send 'remove_from_cart' analytics event 

heitorado:
This PR uses the methods defined on the store sdk related to sending/receiving analytics events, to send the `'remove_from_cart' `event. This event is triggered every time a product is removed from the shopping cart, and then received by an analytics provider that stores the data. The data that is sent is a merge between the fields required by SAE analytics and by GA4, so it can be used by both.

## How it works? 
The `CartItem` component is called at several points in the code, and it uses the 'item' prop passed to it to render the cart item. The item is usually obtained via the `useCart()` (which in turn comes from the `useSDKCart()`), and it contains all the data stored _in the cart_ for that specific item. This means that after an item is added to the cart, its removal will only access the user's cart, and not the item database (which is used to add the item to the cart in the first place). Therefore, the cart stores the item with fewer data than available in the database, so adjustments are necessary to obtain the data that needs to be sent to the analytics. This is implemented as follows:
- When the user adds an item to the cart, the mutate query sent to the server will be triggered with only the allowed fields for the mutation.
- On the client side (indexed DB), however, we will store all the cart attributes (after validated and mutated) + some extra attributes that we'll access later to retrieve the analytics relevant data.
- When the user removes the item from the cart, we check the locally stored data for the item to get all the attributes we need, and trigger the sending of the analytics event.

The remaining process happens as any other event such as in #53 :
The attributes are sent as an analytics event, which then is listened by the AnalyticsHandler and inserted in the `window.dataLayer` array to be consumed by the Analytics service (GA4 or SAE).

## How to test it?
`yarn test`. New tests added on `analytics.test.js`, covering:

- The analytics event is added when clicking on the remove button for any product in the cart
- The `remove_from_cart` analytics event is sent with at least the minimum required attributes.
- To test it manually, add something into the shopping cart and remove it afterwards. Open the console and check if the `window.dataLayer` array has the correct analytics data for the removed item.

## References
remove_from_cart event documentation:
https://developers.google.com/analytics/devguides/collection/ga4/reference/events#remove_from_cart

A really interesting error that may arise in TypeScript and that took a couple of days to solve:
https://stackoverflow.com/a/56701587
https://stackoverflow.com/a/59363875

Last but not least,
@icazevedo, my TypeScript teacher

### DIFF
--- a/cypress/integration/analytics.test.js
+++ b/cypress/integration/analytics.test.js
@@ -112,8 +112,6 @@ describe('remove_from_cart event', () => {
 
           testRemoveFromCartEvent(skuId)
         })
-      cy.getById('cart-item').should('not.exist')
-      cy.getById('checkout-button').should('be.enabled')
     })
   })
 })
--- a/cypress/integration/analytics.test.js
+++ b/cypress/integration/analytics.test.js
@@ -36,39 +36,41 @@ describe('add_to_cart event', () => {
     })
   }
 
-  it('add add_to_cart event in the data layer at product description page', () => {
-    cy.visit(pages.pdp, options)
-    cy.waitForHydration()
+  context('when adding a product to cart', () => {
+    it('adds add_to_cart event in the data layer at product description page', () => {
+      cy.visit(pages.pdp, options)
+      cy.waitForHydration()
 
-    cy.itemsInCart(0)
+      cy.itemsInCart(0)
 
-    // Add to cart
-    cy.getById('buy-button')
-      .click()
-      .then(($btn) => {
-        cy.itemsInCart(1)
-        const skuId = $btn.attr('data-sku')
+      // Add to cart
+      cy.getById('buy-button')
+        .click()
+        .then(($btn) => {
+          cy.itemsInCart(1)
+          const skuId = $btn.attr('data-sku')
 
-        testAddToCartEvent(skuId)
-      })
-  })
+          testAddToCartEvent(skuId)
+        })
+    })
 
-  it('add add_to_cart event in the data layer at the product listing page', () => {
-    cy.visit(pages.collection, options)
-    cy.waitForHydration()
+    it('adds add_to_cart event in the data layer at the product listing page', () => {
+      cy.visit(pages.collection, options)
+      cy.waitForHydration()
 
-    cy.itemsInCart(0)
+      cy.itemsInCart(0)
 
-    // Add to cart
-    cy.getById('buy-button')
-      .first()
-      .click()
-      .then(($btn) => {
-        cy.itemsInCart(1)
-        const skuId = $btn.attr('data-sku')
+      // Add to cart
+      cy.getById('buy-button')
+        .first()
+        .click()
+        .then(($btn) => {
+          cy.itemsInCart(1)
+          const skuId = $btn.attr('data-sku')
 
-        testAddToCartEvent(skuId)
-      })
+          testAddToCartEvent(skuId)
+        })
+    })
   })
 })
 
--- a/cypress/integration/analytics.test.js
+++ b/cypress/integration/analytics.test.js
@@ -14,11 +14,11 @@ const dataLayerHasEvent = (eventName) => {
   })
 }
 
-describe('add_to_cart event', () => {
-  beforeEach(() => {
-    cy.clearIDB()
-  })
+beforeEach(() => {
+  cy.clearIDB()
+})
 
+describe('add_to_cart event', () => {
   const testAddToCartEvent = (skuId) => {
     cy.window().then((window) => {
       const { dataLayer } = window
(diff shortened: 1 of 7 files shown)


cypress/integration/analytics.test.js
```diff
@@ -64,6 +64,52 @@ describe('add_to_cart event', () => {
   })
 })
 
+describe('remove_from_cart event', () => {
```
saranicoly, line 67: Could you add a test to check if the sent values (such as product_reference_id and sku_name) are right?

heitorado: Hmm my answer would be yes, _but_:
- I'm thinking about how could I retrieve those values from the UI (to compare them with what is in the dataLayer). I had to add the `'data-sku'` attribute (at `useRemoveButton.ts`) and I fear that adding more just for the sake of testing will bloat the return of this hook. Thoughts/other solutions?
- Even if those values were available in the UI for comparison, isn't it a bit weird to check for other attributes? (considering that my idea was to make this test only the minimum required things to make it work):
  - Those attributes are optional, and also the dev can at any time choose which attributes to send or not, if they erase some attribute the test would break pointlessly
  - I'm already checking the `sku` attribute not for checking the payload "shape", just for checking that the removed item is the same item in the data layer
  - I'm already checking the required attributes `currency` and `item_name`, and if other dev removes those, the test breaks because the payload is no longer valid.

src/sdk/cart/useRemoveButton.ts
```diff
@@ -17,10 +27,35 @@ export const useRemoveButton = (item: CartItem | null | undefined) => {
         return
       }
 
+      sendAnalyticsEvent<VTEXRemoveFromCartEvent>({
+        type: 'remove_from_cart',
+        data: {
+          currency,
```
icazevedo, line 33:
```suggestion
          currency: code as CurrencyCode,
```

src/sdk/cart/useRemoveButton.ts
```diff
@@ -2,12 +2,22 @@
...
-import type { CartItem } from './validate'
 
-export const useRemoveButton = (item: CartItem | null | undefined) => {
+export const useRemoveButton = (item: AnalyticsCartItem | null | undefined) => {
   const { removeItem } = useCart()
+  const {
+    currency: { code },
+  } = useSession()
+
+  const currency = code as CurrencyCode
```
icazevedo, line 20:
```suggestion
```

icazevedo (approved):
LGTM. Just a couple of nitpicks ⛏️

cypress/integration/analytics.test.js
saranicoly: About how to retrieve the data from UI, you can use `cy.intercept`.  I did a test using it on `analytics.test.js`

https://docs.cypress.io/api/commands/intercept

saranicoly: To me, it makes sense to test other attributes too (even if they are not mandatory) and check if they match the removed product. But this is just a suggestion, I also understand your point, so feel free to ignore it 😄

cypress/integration/analytics.test.js
```diff
@@ -72,6 +72,52 @@ describe('add_to_cart event', () => {
...
+      cy.getById('remove-from-cart-button')
+        .click()
+        .then(($btn) => {
+          cy.itemsInCart(0)
+          const skuId = $btn.attr('data-sku')
+
+          testRemoveFromCartEvent(skuId)
+        })
+      cy.getById('cart-item').should('not.exist')
+      cy.getById('checkout-button').should('be.enabled')
```
eduardoformiga, line 116: I wonder if we should remove these lines since we already have this logic [here in cart.test.js](https://github.com/vtex-sites/base.store/blob/master/cypress/integration/cart.test.js#L87-L88), and the code here is more related to analytics events.

And also, since we don't have any items in the cart, the checkout button shouldn't be enabled, right?

What do you think?

eduardoformiga (approved):
LGTM, just left some comments.

cypress/integration/analytics.test.js
```diff
@@ -72,6 +72,52 @@ describe('add_to_cart event', () => {
...
+
+      const item = event.data.items.find((i) => i.item_variant === skuId)
+
+      expect(item).to.not.be.null
+      expect(item).to.have.property('currency')
+      expect(item).to.have.property('item_name')
+    })
+  }
+
+  context('when removing a product from cart', () => {
```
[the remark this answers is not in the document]
heitorado: Thanks!

cypress/integration/analytics.test.js
heitorado: @eduardoformiga Yeah, makes sense. Better leave these to the specific cart test. Also agree with the checkout button test - not sure why it was passing,  though. I also tested with `cy.getById('checkout-button').should('not.be.enabled')` and it also works, all tests green. Anyways, lines removed. Thanks for the heads up :)

PR #104, opened 2021-11-10
chore: Improve #95

tlgimenes:
There were some issues with #95. This PR fixes:
1. Bug on ProductSummary indices report
2. Remove event computation outside of critical rendering path
3. Replace request interception to use data in the markup instead


cypress.json
```diff
@@ -1,5 +1,5 @@
 {
-  "baseUrl": "http://localhost/",
+  "baseUrl": "http://localhost:9000/",
```
saranicoly, line 2: Isn't the default `localhost:8000` ?

cypress/integration/analytics.test.js
```diff
@@ -147,47 +147,29 @@ describe('view_item event', () => {
...
     cy.getById('product-link')
       .first()
+      .within(() => {
+        cy.getById('buy-button').then(($btn) => {
+          skuId = $btn.attr('data-sku')
+        })
+      })
       .click()
       .then(() => {
-        dataLayerHasEvent('select_item')
```
saranicoly, line 159: Why do not keep this test?

saranicoly (approved):

cypress/integration/analytics.test.js
tlgimenes: it's already taken into account at line 170

cypress.json
tlgimenes: it depends if you are testing with `yarn serve` or `yarn develop`. Anyways, it should be either 9000 or 8000, not like it was before, hehe

saranicoly: Oh, nice

saranicoly (approved):

PR #115, opened 2021-11-19
chore: Simplify search 

tlgimenes:
Simplifies search. More info at: https://github.com/vtex/faststore/pull/1030


igorbrasileiro (approved):
Nice work

src/sdk/search/state.ts
```diff
@@ -0,0 +1,13 @@
+import { navigate } from 'gatsby'
```
igorbrasileiro, line 1: Why can't those functions be inside the framework? Also, the `applySearchState` can receive the navigation callback.

Take a look at this suggestion:
```ts
const applySearchState = (url: URL, navigate: (url: string) => void) => { ... }
```

igorbrasileiro (approved):

src/sdk/search/state.ts
tlgimenes: It's of like this already. The `SearchProvider` accepts the `onChange` callback. This callback returns a URL object instead of the string url.

src/components/sections/ProductGallery/ProductGallery.tsx
```diff
@@ -1,84 +1,83 @@
...
-        >
-          Show More
-        </a>
+        <>
+          <GatsbySeo linkTags={[{ rel: 'next', href: next.link }]} />
+          <a
+            className="p-8 block center"
+            data-testid="show-more"
+            onClick={(e) => {
+              ;(e.target as any).blur?.()
```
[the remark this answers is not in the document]
tlgimenes: Nice! I didn't know about it

PR #118, opened 2021-11-29
feat: Add RC script

igorbrasileiro:
This PR adds the Rquest Capture(RC) script to the store. RC is a VTEX analytics engine.

## How it works? 
The VTEX RC addition works in two steps: 
1. Add the script via <script> tag on the **gatsby-ssr**
2. Add the global variable to push client events on the **gatsby-browser**

## How to test it?
Check if the script is inside the DOM and if the vtexrca has events.

## References
-


gatsby-browser.js
```diff
@@ -33,3 +33,15 @@ export const wrapRootElement = ({ element }) => (
 )
 
 export const wrapPageElement = ({ element }) => <Layout>{element}</Layout>
+
+export const onInitialClientRender = () => {
```
tlgimenes, line 37: We can move this to the global scope I guess

igorbrasileiro: Done

src/typings/global.d.ts
```diff
@@ -1,3 +1,8 @@
 interface Window extends Window {
   dataLayer: any[];
+  vtexrca: {
```
victorhmp, line 3: How about adding some comments here just explaining what these properties mean?

src/sdk/analytics/handlers/RequestCapture.ts
```diff
@@ -0,0 +1,17 @@
+import type { AnalyticsEvent } from '@faststore/sdk'
+
+if (typeof window !== 'undefined') {
+  window.vtexrca =
+    window.vtexrca ||
+    function () {
+      // RCA script
```
victorhmp, line 7: Do we have documentation on this? It would be nice to link it here 😄

src/typings/global.d.ts
igorbrasileiro: Great

victorhmp (approved):

igorbrasileiro:
TODO: Remove the script that creates the vtexrca and put it on RequestCapture.ts. Nowadays exists a problem that the rc.js script tries to create a vtexrca.q.push property, but the vtexrca isn't on Window, because the `requestcapture` script isn't loaded in the browser yet.

igorbrasileiro:
Break send event tasks into new tasks.

PR #137, opened 2021-12-10
[Main Styled] Creates Pattern Library Structure

renatamottam:
`🚨 This page should only be alive during the styling phase.`

Creates a new page to list all the styled components so we can keep the code organized.

|Preview|
|-|
|image|

## How it works? 
Whenever you're going to add a new component to the list, you should:
1. Create a new `grid-section`
2. Update its title
3. Use the `<ul>` to list all the component's variants

### DIFF
--- a/src/styles/pattern-library.scss
+++ b/src/styles/pattern-library.scss
@@ -1,0 +1,21 @@
+// Default //////////////////////////
+
+body {
+  background-color: var(--bg-neutral-light);
+  padding: var(--space-7) 0 var(--space-5);
+}
+
+#gatsby-focus-wrapper > footer,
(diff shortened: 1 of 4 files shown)


src/styles/pattern-library.scss
```diff
@@ -0,0 +1,26 @@
+// Default //////////////////////////
+
+body {
+  padding: var(--space-7) 0 var(--space-5);
+  background-color: var(--bg-neutral-light);
+}
+
+#gatsby-focus-wrapper > header,
```
hellofanny, line 8: We shouldn't do this here. It won't work. I'll find a workaround for it :)

PR #139, opened 2021-12-11
refactor: add image components

tlgimenes:
This PR adds both `<Image/>` and `<ImageWithArtDirection/>` components, both respecting Gatsby Image's naming and props.

## How it works? 
Before, we only provide one hook for handling images. This hook returned and `image` object that we could use to configure `<GatsbyImage/>` component. This made the developer to do a code like:
```tsx
import { GatsbyImage } from 'gatsby-plugin-image'
import { useImage } from '/src/sdk/useImage'

function MyComponent () {
    const image = useImage(...)
    
    return (
      <GatsbyImage image={image} {...otherProps} />
    )
}
```

This adds unnecessary complexity to the code since the `useImage` hook will always be used with the GatsbyImage component.

To address this issue, I'm adding a component that has the hook built in so the developer can write code like:
```tsx
import { Image } from 'src/components/ui/Image'

function MyComponent () {
    return (
      <Image {...imageProps }/>
    )
}
```

## How to test it?
Nothing should have changed, so make sure everything works correctly


igorbrasileiro (approved):
Thank you for the improvement! Great job @tlgimenes

PR #141, opened 2021-12-13
feat: Tiles component

eduardoformiga:
This PR aims to create the basic structure (without the style, we are going to style in other PR afterward) of the `Tiles` component that can be considered a grid with very specific behavior at this moment.

## How it works? 
- The `Tiles` Component can receive from 2 to 4 children.
- Every child should be the `Tile` component.

Basic structure:
```js
<Tiles>
  <Tile>...</Tile>
  <Tile>...</Tile>
  <Tile>...</Tile>
  <Tile>...</Tile>
</Tiles>
```

In the future, we should refactor this component to consider a more flexible way to rearrange each internal `Tile`. So far, the following scenarios are accomplished (desktop/mobile):
- 2 children: equals space.
Screen Shot 2021-12-13 at 19 28 57

- 3 children: First child fits the double of the space. The `Tiles` component internally counts the children and adds a new class just in the first child.
Screen Shot 2021-12-13 at 19 51 34

- 4 children: equals space.
Screen Shot 2021-12-13 at 19 56 05

## How to test it?

## References
https://learncssgrid.com/
https://css-tricks.com/snippets/css/complete-guide-grid/
https://www.origamid.com/projetos/css-grid-layout-guia-completo/


src/components/ui/Tiles/Tiles.tsx
```diff
@@ -0,0 +1,60 @@
...
+  if (isOutOfBounds) {
+    throw new Error(
+      `Tiles cannot receive less than ${MIN_CHILDREN} or more than ${MAX_CHILDREN} children.`
+    )
+  }
+
+  React.Children.forEach(children, (child) => {
+    if (!React.isValidElement(child) || child.type !== Tile) {
+      throw new Error('Only Tile components allowed as children.')
+    }
```
danzanzini, line 36: Is it possible to define the type of the children as an array of Tile so there's no need to check this at runtime?

src/components/ui/Tiles/Tiles.tsx
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  return (
+    <div ref={ref} data-store-tiles data-testid={testId} {...otherProps}>
+      {React.Children.map(children, (child, index) => {
+        if (!React.isValidElement(child)) return
```
danzanzini, line 42: This is already being checked at the conditional above.

src/components/ui/Tiles/Tiles.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+        if (index === 0 && spanFirstChild) {
+          const { className } = child.props
+
+          return React.cloneElement(child, {
+            className: className
+              ? `${className} tile-item-spanded`
+              : 'tile-item-spanded',
+          })
+        }
```
danzanzini, line 52: I may be wrong, but I'd move the content of this conditional to a function to improve the readability of the main return method.

danzanzini (commented):
Good job.

I left some comments on code, but remember that this is my first component review, so I may not be knowing what I'm talking about 😄

src/components/ui/Tiles/Tiles.tsx
eduardoformiga: I tried to do that and also have read some [threads](https://github.com/microsoft/TypeScript/issues/13618#issuecomment-307701505) about that. But it looks like we can't enforce to just have Tile's Array as children.

src/components/ui/Tiles/Tiles.tsx
eduardoformiga: Agree with you. 
Just kept this line because `eslint` complains if we didn't check it again. In this case, I prefer to keep than add some `eslint` ignore rules.

UquutCuABM

src/components/ui/Tiles/Tiles.tsx
eduardoformiga: Hmm, not sure. This ternary operator is so common in React when we tweak `className`.

eduardoformiga:
> Good job.
> 
> I left some comments on code, but remember that this is my first component review, so I may not be knowing what I'm talking about 😄

Hey, No worries. It's always good to receive reviews and improve our codebase and skills. 😄

src/components/ui/Tiles/Tiles.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <div ref={ref} data-store-tiles data-testid={testId} {...otherProps}>
+      {React.Children.map(children as ReactElement, (child, index) => {
+        if (index === 0 && spanFirstChild) {
+          const { className } = child.props
+
+          return React.cloneElement(child, {
+            className: className
+              ? `${className} tile-item-spanded`
+              : 'tile-item-spanded',
+          })
```
filipewl, line 49: 🆒

filipewl (approved):
Nice Forms! Learned some things.

danzanzini (approved):

src/components/ui/Tiles/Tiles.tsx
tlgimenes: At least move this check inside the `map` statement so we don't loop twice over the children

src/components/ui/Tiles/Tiles.tsx
```diff
@@ -0,0 +1,58 @@
...
+  if (isOutOfBounds) {
+    throw new Error(
+      `Tiles cannot receive less than ${MIN_CHILDREN} or more than ${MAX_CHILDREN} children.`
+    )
+  }
+
+  React.Children.forEach(children as ReactElement, (child) => {
+    if (child.type !== Tile) {
+      throw new Error('Only Tile components allowed as children.')
+    }
```
igorbrasileiro, line 36: Does this error is being used to warn the developer about the bounds? What do you think about only warning on developing mode?
```suggestion
    if (process.env.NODE_ENV === 'development') {
      if (child.type !== Tile) {
        throw new error ('Only Tile components allowed as children.')
      }
    }
```

src/components/ui/Tiles/Tiles.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const childrenCount = React.Children.count(children)
+  const spanFirstChild = childrenCount === NUMBER_ITEMS_TO_EXPAND
+  const isOutOfBounds =
+    childrenCount < MIN_CHILDREN || childrenCount > MAX_CHILDREN
+
+  if (isOutOfBounds) {
+    throw new Error(
+      `Tiles cannot receive less than ${MIN_CHILDREN} or more than ${MAX_CHILDREN} children.`
+    )
+  }
```
igorbrasileiro, line 31: Does this error is being used to warn the developer about the bounds? What do you think about only warning on developing mode?

```suggestion
  if (process.env.NODE_ENV === 'development') {
    if (isOutOfBounds) {
      throw new Error(
        `Tiles cannot receive less than ${MIN_CHILDREN} or more than ${MAX_CHILDREN} children.`
      )
    }
  }
```

src/components/ui/Tiles/Tiles.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <div ref={ref} data-store-tiles data-testid={testId} {...otherProps}>
+      {React.Children.map(children as ReactElement, (child, index) => {
+        if (index === 0 && spanFirstChild) {
+          const { className } = child.props
+
+          return React.cloneElement(child, {
+            className: className
+              ? `${className} tile-item-spanded`
```
igorbrasileiro, line 47: I'm my opinion the children should be opaque for the parent. You can implement this using another approach via CSS.

```tsx
return (
  <div {...otherProps} data-store-tiles={spanFirstChild ? 'expanded' : 'normal'}>{children}</div>
)
```

on the CSS file you can do this:
```css
[data-store-tiles] {
  /* whatever CSS you wan on both cases normal or expanded */
}

[data-store-tiles= "expanded"] > [data-store-tile]:first-child {
  /* whatever CSS you want to apply to the first data-store-tile element **when** data-store-tiles is expanded */
}
```

igorbrasileiro: It will make your code a little bit faster and easier to read.

igorbrasileiro (changes requested):

src/components/ui/Tiles/Tiles.tsx
eduardoformiga: Heey, I really liked this approach! I'm gonna implement it, thanks!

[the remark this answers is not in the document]
eduardoformiga: It isn't a tailwind class. 
I'm gonna follow @igorbrasileiro's [suggestion below](https://github.com/vtex-sites/base.store/pull/141#discussion_r772650759). So we can avoid this clone operation :)

src/components/ui/Tiles/Tiles.tsx
eduardoformiga: Good suggestion, thanks. 
I'm gonna change the implementation (remove the `map`) so that we just need to loop once to check the `child`'s type.

src/components/ui/Tiles/Tiles.tsx
eduardoformiga: Hmm, not sure. Since we don't want to skip the validation in the other environments like `staging` or `production`.
I also think is a little weird to check the environment inside the component. What do you think?

src/components/ui/Tiles/Tiles.tsx
eduardoformiga: We'd like to ensure at this moment that we just have from 2 to 4 children as `Tile`. I think the [same comment above](https://github.com/vtex-sites/base.store/pull/141#discussion_r772730022) applies here. What do you think?

src/components/ui/Tiles/Tiles.tsx
danzanzini: I think environment checks should be the closest as possible to the request handler.

The component raises the error, whoever calls this component should be responsible to handle (or raise) the error and, if necessary, applying some environment check.

danzanzini (approved):
Great Job, @eduardoformiga. I'm approving the PR but have one last question:

I think we have the tiles being used at the [alpha homepage](https://www.figma.com/file/cAXtiQGObWxamFVVhviQVM/Handoff-Homepage-v0.2?node-id=901%3A58296). Are you going to apply it now or only at the styling moment?

src/components/ui/Tiles/Tiles.tsx
igorbrasileiro: Some references : [mui](https://github.com/mui-org/material-ui/blob/1230b1a571942f1abf2fec81812b757ceb0e5d0b/packages/mui-system/src/spacing.js#L104), [reakit](https://github.com/reakit/reakit/blob/a211d94da9f3b683182568a56479b91afb1b85ae/packages/reakit-warning/src/warning.ts#L13).

igorbrasileiro (approved):
Great job, @eduardoformiga.
I still think it would be good to remove those checks from the production build.

eduardoformiga:
> Great Job, @eduardoformiga. I'm approving the PR but have one last question:
> 
> I think we have the tiles being used at the [alpha homepage](https://www.figma.com/file/cAXtiQGObWxamFVVhviQVM/Handoff-Homepage-v0.2?node-id=901%3A58296). Are you going to apply it now or only at the styling moment?

As we talked we're going to implement this in other tasks afterward ;)

PR #142, opened 2021-12-14
feat(ui): Add image gallery component

saranicoly:
<em>The [image gallery](https://www.figma.com/file/zHMOtOPduiM0hNNcg21iR1/Handoff-PDP-v0.1?node-id=18%3A33225) component shows all the product images on PDP.

Screen Shot 2021-12-23 at 21 16 49

It's currently hidden from PDP because it won't be used right now.
</em>

## How it works? 
<em>To use ImageGallery, you must import it from `src/components/ui/ImageGallery` and call the component like this:
`<ImageGallery images={imagesArray}/>`

**imagesArray** is an array of `ImageElementData`:
```TypeScript
export interface ImageElementData {
  url: string
  alternateName: string
}
```
Please note that using Thumbor generated images will not work as expected. Use simple image urls if you want to create this array of objects manually. Good image sources: 
- [Placekitten](https://placekitten.com/)
- [Pexels](https://www.pexels.com/)
</em>

## How to test it?
<em> 

- Import ImageGallery in `ProductDetails.tsx` and change Image tag for: `<ImageGallery images={productImages} />`. You can also hardcode your own array of `ImageElementData` using your own image urls.
- run `yarn develop` and `yarn test`
</em>


israelswf:
Just a minor suggestion I've notice now when following the _How to_ steps ... change:
`<ImageGallery images={all_images}/>`
to
`<ImageGallery images={allImages}/>`

Eslint is complaining about the `all_images` variable name in _snake case_.

src/components/ui/ImageGallery/ImageGalleryZoom.tsx
```diff
@@ -0,0 +1,15 @@
...
+  zoomFactor?: number
+}
+
+const ImageGalleryZoom = ({
+  children,
+}: PropsWithChildren<ImageGalleryZoomProps>) => {
+  return <> {children} </>
+}
+
+export default ImageGalleryZoom
```
igorbrasileiro, line 15: Remove this noop component

igorbrasileiro (changes requested):
Great job, @saranicoly! I left some comments.
I have a few questions, why need to reimplement the whole carousel?
Also, what do you think about this structure?

```tsx
const ImageGallery = ({ images }) => {
   const [selectedImage, setSelectedImage] = useState(0)
   const image = images[selectedImage]
   
   <div>
      <ZoomImage>
         <Image {...image} />
      </ZoomImage>
      <Carousel>
        {images.map((image, index) => {
          return <Image {...image} onClick={(e) => { setSelectedImage(index) })
        })}
      </Carousel>
   </div>
}
```

Maybe I'm forgetting something, but IMO it will make this component a little bit easier to understand.

lariciamota:
Hi! I wasn't able to open the figma link :/
Screen Shot 2021-12-22 at 16 52 24

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,139 @@
...
+        aria-label="slide image selector forward"
+        onClick={() => {
+          if (sliderState.sliding) {
+            return
+          }
+
+          slide('next', sliderDispatch)
+        }}
+        icon={<ForwardArrowIcon color="#323845" />}
+      />
```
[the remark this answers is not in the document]
saranicoly: Done. Thanks 😊

src/components/ui/ImageGallery/ImageGalleryProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+  onClick,
+  ...otherProps
+}: ImageGalleryProps) => {
+  const contextValues = useMemo(() => {
+    return { name, selectedImageData, onClick }
+  }, [name, selectedImageData, onClick])
+
+  return (
+    <ImageGalleryContext.Provider value={contextValues}>
+      <div {...otherProps}>{children}</div>
```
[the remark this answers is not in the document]
heitorado: We adopted your previous suggestion of structure, so `ImageGalleryContext.Provider` does not exist anymore.

src/components/ui/ImageGallery/ImageGallery.tsx
```diff
@@ -0,0 +1,48 @@
...
+        setCurrentImage((previousSelectedImage) => {
+          const target = v.target as HTMLImageElement
+
+          return {
+            url: target.getAttribute('src') ?? previousSelectedImage.url,
+            alternateName:
+              target.getAttribute('alt') ?? previousSelectedImage.alternateName,
+          }
+        })
+      }
```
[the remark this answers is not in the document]
heitorado: This initial design was putting the `onClick` in the button so the `e.target` would be the image component, from where we accessed the `src` and `alt` parameters. Now after the rewrite on the structure, the `onClick` is on the button that wraps the image component and does not need to grab attributes from its child anymore. The main reason that we did not put the `onClick` directly into the image, but on the Button instead, is to give better semantics.

We thought about adding the `button` role, but we found [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) that some extra work is necessary to make it _behave_ like a proper button. Thus, we decided to stick with the Button component instead.

src/components/ui/ImageGallery/ImageGalleryZoom.tsx
saranicoly: The idea of having this component is just to be a placeholder with some basic structure, so it’s going to be ready to receive its functionality later. With that in mind, we think this should be in a separated file like this because it's not being used right now but it will in the future.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,139 @@
...
+  for (let idx = 0; idx < totalItems; ++idx) {
+    const transformValue = -(slideWidth * idx)
+
+    transformMap[idx] = transformValue
+  }
+
+  return transformMap
+}
+
+function ImageGallerySelector({ imagesPerPage, images, ...otherProps }: Props) {
```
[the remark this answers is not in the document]
heitorado: I agree. This is, in fact, an adaptation of the shelf implementation made by @georgebrindeiro to the faststore 101 class. We thought about renaming it to "Slider" but we already have a component with that name. Any thoughts/suggestions on this?

src/components/ui/ImageGallery/ImageGallerySelector.tsx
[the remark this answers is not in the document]
heitorado: Fixed at https://github.com/vtex-sites/base.store/pull/142/commits/a9e1b45dfc591c5d5364d46eda79c9f6094fb61c

src/components/ui/ImageGallery/ImageGalleryZoom.tsx
saranicoly: Also, the zoom component has value itself. So maybe it could be reused in other structures.

heitorado:
> I have a few questions, why need to reimplement the whole carousel?

Hmm, it wasn't our intention to reimplement the carousel. In fact, initially we tried to use the Carousel component, but we perceived that:
- It isn't flexible about the number of elements to be displayed on a page
- We tried to update the component to add that flexibility, but apparently there is a lot of internal logic that relies on displaying a single item per page. We found that it was easier to work with the basic shelf implementation than the Carousel.
- Also, we would have to add the flexibility to receive custom arrow icons and the amount of items per page.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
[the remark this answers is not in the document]
saranicoly: What do you think about `ImageSelector` or just `Selector`?

saranicoly:
> Hi! I wasn't able to open the figma link :/ Screen Shot 2021-12-22 at 16 52 24

Sorry, I think you don't have access. 😢 
We will change the PR description and put an image instead of the Figma link so that everyone can see it.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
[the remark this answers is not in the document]
igorbrasileiro: I prefer to name it a carousel because the slider is a kind of input.

igorbrasileiro (approved):
Great job, @saranicoly, and @heitorado!

src/components/ui/ImageGallery/ImageGalleryZoom.tsx
igorbrasileiro: Ok, no problem.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+  itemsPerPage: number
+}
+
+const createTransformValues = (totalItems: number) => {
+  if (!totalItems) {
+    return {}
+  }
+
+  const transformMap: Record<number, number> = {}
+  const slideWidth = 100 / totalItems
```
eduardoformiga, line 17: What does this 100 magic number represent? the entire width?

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  const { handlers, slide, sliderState, sliderDispatch } = useSlider({
+    totalItems: elementCount,
+    itemsPerPage,
+    infiniteMode: false,
+  })
+
+  const slidingTransition = `transform 400ms`
+
+  if (!elements || !elementCount) {
```
eduardoformiga, line 49:
```suggestion
  if (!elementsCount) {
```
I think the first condition (![]) is always false since the array is an object and is always evaluated as a [truthy](https://developer.mozilla.org/pt-BR/docs/Glossary/Truthy) value, but with the `!` operator will be always `falsy`. 

I also wonder if we should move this guard clause to the beginning of the function.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+      />
+
+      <div {...otherProps} {...handlers}>
+        <div
+          data-carousel-track
+          style={{
+            display: 'flex',
+            transition: sliderState.sliding ? slidingTransition : undefined,
+            width: `${(elementCount * 100) / itemsPerPage}%`,
+            transform: `translate3d(${
```
eduardoformiga, line 85: 🆒 !

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+          slide('next', sliderDispatch)
+        }}
+      />
+
+      <div {...otherProps} {...handlers}>
+        <div
+          data-carousel-track
+          style={{
+            display: 'flex',
```
eduardoformiga, line 82: Nice job here!
I think you have a plan to move the static CSS (the ones that don't depend on js values) in the style PR afterward, right?

eduardoformiga (commented):
Amazing job, folks! ✨  
I Left some comments.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
saranicoly: Yes, I think so.

src/components/sections/ProductDetails/ProductDetails.tsx
```diff
@@ -30,7 +30,7 @@ function ProductDetails({ product: staleProduct }: Props) {
       name: variantName,
       brand: { name: brandName },
       isVariantOf: { name, productGroupID: productId },
-      image: [img],
+      image: productImages,
```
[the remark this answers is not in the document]
heitorado: When changing this, my initial thought was that it would be "one less thing to do" when we integrate the gallery to the PDP. Just pass the "productImages" to the component and it's done.

src/components/ui/ImageGallery/ImageGallery.tsx
```diff
@@ -0,0 +1,58 @@
...
+        />
+      </ImageZoom>
+      <ImageGallerySelector itemsPerPage={4}>
+        {images.map((image, idx) => {
+          return (
+            <Button
+              key={idx}
+              data-thumbnail-button
+              className={image.url === currentImage.url ? 'selected' : ''}
+              aria-label={`${image.alternateName} - Image ${idx + 1} of ${
```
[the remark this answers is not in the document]
heitorado: I'm not sure if we should completely remove it. One is for the image being currently displayed, and other is for the images in the selector. Since we have two different "contexts", I guess it would make sense for the screen reader to describe the image that is currently appearing and the list of images that can be clicked to switch the main image? I do not know much about screen readers as well.

I used this site as a reference:
https://www.rei.com/product/194789/teva-reember-slip-ons-womens

It is listed on [baymard](https://baymard.com/blog/accessibility-benchmark-launch) as having good accessibility (just saying this to explain that it is not a random store):

Also just noticed that I made a mistake - the main (selected) image does not have the second part of the alt (after the `-`). Maybe it would be better if we strip the "- Image x of y" part from the selected image, as it is in the aforementioned store?

src/components/ui/ImageGallery/ImageGallerySelector.tsx
heitorado: Credits to @georgebrindeiro

src/components/ui/ImageGallery/ImageGallerySelector.tsx
heitorado: If we try to move the guard clause to the beginning of the function, this error is reported from the linter:

> _React Hook "useSlider" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?_
(similar error arises for `useMemo`)

https://reactjs.org/docs/hooks-rules.html

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+    [elementCount]
+  )
+
+  const { handlers, slide, sliderState, sliderDispatch } = useSlider({
+    totalItems: elementCount,
+    itemsPerPage,
+    infiniteMode: false,
+  })
+
+  const slidingTransition = `transform 400ms`
```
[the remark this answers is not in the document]
heitorado: Done at https://github.com/vtex-sites/base.store/pull/142/commits/6507a1eebf04482dcb0ca2e91a24fd525a6effab

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  if (!elements || !elementCount) {
+    return null
+  }
+
+  return (
+    <section>
```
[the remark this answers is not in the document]
heitorado: Done at https://github.com/vtex-sites/base.store/pull/142/commits/d20395480eb39d162befb881bcf8add423298866

src/components/ui/ImageGallery/ImageGallery.tsx
```diff
@@ -0,0 +1,58 @@
...
+              }`}
+              onClick={() => {
+                setSelectedImageIdx(idx)
+              }}
+            >
+              <Image
+                src={image.url}
+                alt={image.alternateName}
+                variant="product.details"
+                loading="eager"
```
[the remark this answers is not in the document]
heitorado: Good call. Done at https://github.com/vtex-sites/base.store/pull/142/commits/9091cff7757c18b78f22327472f751232117a1d6

src/components/ui/ImageGallery/ImageGallery.tsx
```diff
@@ -0,0 +1,58 @@
...
+          loading="eager"
+        />
+      </ImageZoom>
+      <ImageGallerySelector itemsPerPage={4}>
+        {images.map((image, idx) => {
+          return (
+            <Button
+              key={idx}
+              data-thumbnail-button
+              className={image.url === currentImage.url ? 'selected' : ''}
```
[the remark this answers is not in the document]
heitorado: Hmm, I don't know. I'd rather stay with the 'selected' in the classname rather than using the data attribute for that. Personally I'd prefer doing `[data-thumbnail-button].selected` instead of `[data-thumbnail-button='selected']`. On both cases, the 'default' button state would be styled with just `[data-thumbnail-button]`, right?

Not sure which way is better. @renatamottam, @saranicoly - thoughts on this?

src/components/ui/ImageGallery/ImageGallerySelector.tsx
heitorado: About that:

> I think the first condition (![]) is always false since the array is an object and is always evaluated as a [truthy](https://developer.mozilla.org/pt-BR/docs/Glossary/Truthy) value, but with the `!` operator will be always `falsy`.

You are right. I thought that if the children happened to be `[]`, `null` or `undefined`, the `.toArray` method would behave unexpectedly, but it always returns an empty array. And like you said, `![]` is always `false`. Fixed at

src/components/ui/ImageGallery/ImageGallery.tsx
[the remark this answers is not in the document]
eduardoformiga: Good references 😄 

I said that because we have the `alternateName` in three different places:
- In the `Image` inside the `ImageZoom` as `alt`.
- In the `Button` inside `ImageGallerySelector` as `aria-label`.
- In the `Image` inside the `Button` as `alt`.

It could sound a little annoying for screen readers. But no worries, good job worrying about it!

eduardoformiga (approved):

src/components/ui/ImageGallery/ImageGallery.tsx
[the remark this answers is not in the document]
hellofanny: Let's go with igor's suggestion! Can you please use `[data-thumbnail-button='selected']` instead? We are trying to follow this pattern for the other components too :)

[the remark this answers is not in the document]
heitorado: Great! So data attribute it is. Thanks, @igorbrasileiro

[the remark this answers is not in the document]
heitorado: Done at https://github.com/vtex-sites/base.store/pull/142/commits/11880b496ee56279d428dca98d447ac477542b58

src/components/ui/ImageGallery/ImageGallery.tsx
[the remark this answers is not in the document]
heitorado: Hmm got it. Just now I realized that we have it three times, in which two of these times is the same message. Not sure how can we improve it without sacrificing one `alt` or `aria-label`. Suggestions? @saranicoly, thoughts on this?

src/components/ui/ImageGallery/ImageGallerySelector.tsx
```diff
@@ -0,0 +1,128 @@
...
+        onClick={() => {
+          if (sliderState.sliding) {
+            return
+          }
+
+          slide('previous', sliderDispatch)
+        }}
+      />
+      <IconButton
+        aria-label="slide image selector forward"
```
lariciamota, line 65: What do you think about this suggestion?
I'm not 100% sure about the English, but for me putting it at the beginning sounds better and starts with the most important part of the component description
```suggestion
        aria-label="backward slide image selector"
        icon={<BackwardArrowIcon color="#323845" />}
        onClick={() => {
          if (sliderState.sliding) {
            return
          }

          slide('previous', sliderDispatch)
        }}
      />
      <IconButton
        aria-label="forward slide image selector"
```

src/components/ui/ImageGallery/ImageGallerySelector.tsx
lariciamota: I'm curious, why will it be done separately?

src/components/ui/ImageGallery/ImageZoom.tsx
```diff
@@ -0,0 +1,13 @@
+import type { PropsWithChildren } from 'react'
+import React from 'react'
+
+interface ImageZoomProps {
+  helpMessage?: string
+  zoomFactor?: number
+}
+
+const ImageZoom = ({ children }: PropsWithChildren<ImageZoomProps>) => {
+  return <> {children} </>
```
igorbrasileiro, line 10:
```suggestion
  return children
```

src/components/ui/ImageGallery/ImageGallerySelector.tsx
igorbrasileiro: Something interesting, why not calculate only the current transformValue instead of all values? I am asking because I wanna remove the useMemo, if it's possible.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
heitorado: Yes, it is the entire width of the element on which the ImageGallery is nested. I'll add a comment to make that explicit.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
heitorado: Great suggestion. I believe the English is also fine.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
heitorado: Done at https://github.com/vtex-sites/base.store/pull/142/commits/8f303c5e398e46c7c0396bafbdd32aec813322aa

src/components/ui/ImageGallery/ImageGallerySelector.tsx
heitorado: The ImageGallerySelector will not be present in the alpha version of the styled base.store. Since we are focusing on different component integrations/styling now, the styling for this component will be on hold for a while.

src/components/ui/ImageGallery/ImageGallerySelector.tsx
heitorado: Good call. I've removed the `useMemo` here: https://github.com/vtex-sites/base.store/pull/142/commits/d244c4dec6b48a0c4a97f8f7379f88285bd8739d

src/components/ui/ImageGallery/ImageZoom.tsx
heitorado: I tried adding your change, but then I got the following error at `ImageGallery.tsx` when using the `<ImageZoom />`:

```
'ImageZoom' cannot be used as a JSX component.
  Its return type 'ReactNode' is not a valid JSX element.
    Type 'undefined' is not assignable to type 'Element | null'
```

PR #143, opened 2021-12-14
Add icons

gvc:
In order to properly style the new base.store UI, we're adding the Phosphor icons react library, to see if it impacts the overall lighthouse score. The icons being used in this test will be used in the final version


gvc:
I've built `yarn build` and checked the final bundle and it tree-shaked, so I think we should be fine. To do this I copied the path of the used icons on this PR and checked to see if they were on the final payload. I then proceeded to select a few that weren't used and checked that they **weren't** on the final payload.

eduardoformiga (approved):
Icons ✅ 
tree-shaking ✅ 
Forms review ✅ 🎉

PR #146, opened 2021-12-14
feat(ui): Add Out of Stock component

israelswf:
This PR will add the basic structure for the `Out of Stock` component (it will be styled in a future PR), that can be used 
to keep users informed when a product is out of stock. 
image
 
TODO (in a future PR, due to low prioritization of the component)
- Add icon to button
-  Display success alert if onSubmit succeed
- Handle Aria live Regions

## How it works? 
This component is going to be displayed in the PDP whenever a product is out of stock. The component will receive a callback function that will be trigged inside it, taking the email (filled by the user) as a parameter. 

## How to test it?
- By running the PDP page (at the moment it is always being displayed, regarding the availability of the product)
- `Yarn test`

## References
[ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)


src/components/product/OutOfStock/OutOfStock.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  const [email, setEmail] = useState('')
+
+  // TODO: Display (after the button) a success alert if onSubmit succeed
+
+  return (
+    <>
+      <p>{title}</p>
+      <Label>{notificationMsg}</Label>
+      <Input
```
filipewl, line 27: Isra, do remember to connect the input and its label via an `id` & `htmlFor`.

src/components/sections/ProductDetails/ProductDetails.tsx
```diff
@@ -70,6 +71,7 @@ function ProductDetails({ product: staleProduct }: Props) {
       <Button {...buyProps} disabled={isValidating}>
         Add to cart
       </Button>
+      <OutOfStock onSubmit={(email) => alert(email)} />
```
filipewl, line 74: [Off-topic] Food for later thought: understand how [ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) work. I feel like this component is a case for this because it only appears as the user is interacting with the SKU selector and needs (?) to be announced by assistive technologies.

israelswf: Thanks for sharing it, @filipewl. It is a live region, indeed. I guess it is something we are only going to tackle in the future (due to our priorities), but I'm to leave comments in the code about it.

src/components/product/OutOfStock/OutOfStock.tsx
```diff
@@ -0,0 +1,62 @@
...
+    notificationMsg = 'Notify me when available',
+    buttonTxt = 'Send',
+    icon = defaultIcon,
+    onSubmit,
+  } = props
+
+  const [email, setEmail] = useState('')
+
+  return (
+    <>
```
lucasfp13, line 42: I'm wondering if wouldn't be better to use a `div` here and fill it alongside the other HTML tags with the standard data attributes we're used to add for each component (`data-testid`, `data-store-...`)

lucasfp13 (commented):
I think there are some things missing. After review that, LGTM

filipewl (approved):

PR #147, opened 2021-12-14
Temporarily fix affordances on links and buttons

gvc:
After removing the small amount of CSS we had, we incurred in some lighthouse failures. This is a temporary fix for it, until we actually style the buttons and links.

### DIFF
--- a/src/styles/typography.scss
+++ b/src/styles/typography.scss
@@ -91,3 +91,11 @@ h1, h2, h3, h4, h5, h6, p, ul {
   font-size: var(--text-size-3);
   line-height: 1.5;
 }
+
+/* FIXME Remove this once links and buttons are styled properly */
+a, button[data-store-button] {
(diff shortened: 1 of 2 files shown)


src/styles/typography.scss
```diff
@@ -91,3 +91,11 @@ h1, h2, h3, h4, h5, h6, p, ul {
   font-size: var(--text-size-3);
   line-height: 1.5;
 }
+
+/* FIXME Remove this once links and buttons are styled properly */
+a, button[data-store-button] {
```
hellofanny, line 96: Hey @gvc, Thank for taking care of this! 
Since it's temporary, can you please move this to `pattern-library.scss` file?

gvc: Will do this as soon as Lighthouse indicates I'm in the right direction :)

PR #150, opened 2021-12-16
Feat: style SearchInput component

hellofanny:
- Style SearchInput Component

|Default|
|-|
default

|Focus|
|-|
focus

## How it works? 
<em>Tell us the role of the new feature, or component, in its context.</em>

## How to test it?

## References
<em>Spread the knowledge: is this any content you used to create this PR that is worth sharing?</em>  

<em>Extra tip: add references to related issues or mention people important to this PR may be good for the documentation and reviewing process</em>


src/pages/pattern-library.tsx
```diff
@@ -66,6 +68,15 @@ function Page() {
               </li>
             </ul>
           </section>
+
+          <section className="grid-section grid-content">
+            <h2 className="title-subsection">Inputs</h2>
+            <ul className="list-horizontal">
+              <li style={{ width: '50%' }}>
```
danzanzini, line 75: I'd remove this width definition so we can see at the pattern library that the default behavior of the `SearchInput` component is to fill the whole width.

danzanzini (approved):
Good job @hellofanny. Approving

src/pages/pattern-library.tsx
hellofanny: Actually, in this case the search input width depends on the parent :/ (It will fits its parent width, so I will just change it for 100% for now). Thanks for reviewing :)

PR #151, opened 2021-12-16
Feat: style Alert component

eduardoformiga:
This PR aims to add some variations of the Alert Component in the Pattern Library.
- Default
- Without Icon
- With Dismiss Button
- With Action
 
 ** Using the state just to show the alert after closing it.

Z184fyygbA

## How it works? 
the Alert can receive some optional props like:
- icon: ReactNode that represents the icon (you can pass a [Phosphor Icon](https://phosphoricons.com/)).
- dismissible: boolean that controls the dismissible (close) option.
- onClose: callback function after clicking on the close button.

## How to test it?
http://localhost:8000/pattern-library/#alerts

## References
https://phosphoricons.com/


src/components/ui/Alert/alert.scss
```diff
@@ -0,0 +1,32 @@
+[data-store-alert] {
+  box-sizing: border-box; /** FIXME remove after add css reset */
```
hellofanny, line 2: css reset added :)

src/components/ui/Alert/alert.scss
```diff
@@ -0,0 +1,32 @@
...
+  background-color: var(--bg-primary-light);
+  color: var(--color-text);
+  font-size: var(--text-size-1);
+
+  [data-store-icon] {
+    margin: 0;
+    padding-right: var(--space-1);
+  }
+
+  span { font-weight: bold; }
```
hellofanny, line 16:
```suggestion
  span { font-weight: var(--text-weight-bold); }
```

src/components/ui/Alert/alert.scss
```diff
@@ -0,0 +1,32 @@
...
+    color: var(--color-text);
+
+    &:visited { color: var(--color-text) }
+  }
+
+  [data-alert-button] {
+    margin-left: auto;
+    padding: 0 var(--space-0);
+    border: none;
+    background-image: none;
```
hellofanny, line 29: Not sure if we need this.

src/components/ui/Alert/alert.scss
```diff
@@ -0,0 +1,32 @@
...
+  span { font-weight: bold; }
+
+  a {
+    font-weight: bold;
+    color: var(--color-text);
+
+    &:visited { color: var(--color-text) }
+  }
+
+  [data-alert-button] {
```
hellofanny, line 25: Btw, I'm not sure if we are going to keep the `button[data-store-button]`  width/height: `48px`. If not, we might need to add width and height to this button too.

src/components/ui/Alert/alert.scss
```diff
@@ -0,0 +1,32 @@
...
+
+  [data-store-icon] {
+    margin: 0;
+    padding-right: var(--space-1);
+  }
+
+  span { font-weight: bold; }
+
+  a {
+    font-weight: bold;
```
hellofanny, line 19:
```suggestion
    font-weight: var(--text-weight-bold);
```

hellofanny (approved):
Uuhulll! Nice!!! I finally see some colors 🥳

src/components/ui/Alert/alert.scss
eduardoformiga: Nice catch :)

src/components/ui/Alert/alert.scss
eduardoformiga: That's true. 
When I added the `min-width/min-height: 48px (var(--space-7))` the CSS was broken. But we can fix using.

```css
[data-store-alert] {
  position: relative;
  
  [data-alert-button] {
    position: absolute;
    right: 0;
  }
}
```

PR #159, opened 2021-12-20
feat: Add cart button styles

ArthurTriis1:
This PR styles the cart-toggle component

## How it works? 
Add a stylesheet for cart-toggle and implement a style like the [Figma Prototype](https://www.figma.com/file/wS0Kek1aSzIRl20jdtnSsL/Alpha-Version?node-id=0%3A1)

Screen Shot 2021-12-21 at 12 49 25

### DIFF
--- a/src/components/cart/CartToggle/CartToggle.tsx
+++ b/src/components/cart/CartToggle/CartToggle.tsx
@@ -2,13 +2,13 @@ import React from 'react'
 import Button from 'src/components/ui/Button'
 import { useCartToggleButton } from 'src/sdk/cart/useCartToggleButton'
 
-import * as styles from './cart-toggle.module.scss'
+import './cart-toggle.scss'
 
 function CartToggle() {
   const btnProps = useCartToggleButton()
 
   return (
-    <Button {...btnProps} className={styles.buttonCart}>
+    <Button {...btnProps} data-store-button-cart>
       <img src="https://cdn-icons-png.flaticon.com/512/263/263142.png" alt="" />
     </Button>
   )
--- a/src/components/cart/CartToggle/CartToggle.tsx
+++ b/src/components/cart/CartToggle/CartToggle.tsx
@@ -12,9 +12,9 @@ function CartToggle() {
     <Button
       {...btnProps}
       className="cart-toggle"
-      aria-label="Button cart toggle"
+      aria-label={`Cart with ${btnProps['data-items']} items`}
     >
-      <ShoppingCart />
+      <ShoppingCart size={32} />
     </Button>
   )
 }
--- a/src/components/cart/CartToggle/CartToggle.tsx
+++ b/src/components/cart/CartToggle/CartToggle.tsx
@@ -9,7 +9,11 @@ function CartToggle() {
   const btnProps = useCartToggleButton()
 
   return (
-    <Button {...btnProps} className="cart-toggle">
+    <Button
+      {...btnProps}
+      className="cart-toggle"
+      aria-label="Button cart toggle"
+    >
       <ShoppingCart />
     </Button>
   )
--- a/src/components/cart/CartToggle/CartToggle.tsx
+++ b/src/components/cart/CartToggle/CartToggle.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import Button from 'src/components/ui/Button'
 import { useCartToggleButton } from 'src/sdk/cart/useCartToggleButton'
+import { ShoppingCart } from 'phosphor-react'
 
 import './cart-toggle.scss'
 
@@ -8,8 +9,8 @@ function CartToggle() {
   const btnProps = useCartToggleButton()
 
   return (
-    <Button {...btnProps} data-store-button-cart>
-      <img src="https://cdn-icons-png.flaticon.com/512/263/263142.png" alt="" />
+    <Button {...btnProps} className="cart-toggle">
+      <ShoppingCart />
     </Button>
   )
 }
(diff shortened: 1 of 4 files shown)


src/pages/pattern-library.tsx
```diff
@@ -145,6 +147,19 @@ function Page() {
...
+          <section className="grid-section grid-content">
+            <h2 className="title-subsection">CartToggle</h2>
+            <UIProvider>
+              <CartProvider>
+                <ul className="list-vertical">
+                  <li>
+                    <CartToggle />
+                  </li>
+                </ul>
+              </CartProvider>
```
danzanzini, line 160: This is not a blocker but do we need these SDK providers at the pattern library? Maybe it's better to use the component without integrations here.

danzanzini (approved):
LGTM 🚀

src/pages/pattern-library.tsx
ArthurTriis1: The component  use contexts, then we need to put their inside a provider

src/components/cart/CartToggle/cart-toggle.scss
```diff
@@ -0,0 +1,39 @@
...
+      top: var(--space-0);
+      left: var(--space-4);
+
+      font-weight: var(--text-weight-bold);
+      font-size: var(--text-size-0);
+      color: var(--color-text-negative);
+
+      background: var(--bg-secondary-default);
+      border-radius: var(--border-radius-pill);
+      padding: calc(var(--space-0)/4) var(--space-0);
```
renatamottam, line 26: I think we can create a new variable for `1px` or just use `1px` here.

src/components/cart/CartToggle/cart-toggle.scss
```diff
@@ -0,0 +1,39 @@
+[data-store-button].cart-toggle {
+    cursor: pointer;
+
+    background-color: #FFF0;
```
renatamottam, line 4: We should always use our color variables.

src/components/cart/CartToggle/CartToggle.tsx
```diff
@@ -1,15 +1,20 @@
...
   const btnProps = useCartToggleButton()
-  const { totalUniqueItems } = useCart()
 
   return (
-    <Button {...btnProps} className="mx-3">
-      cart: {totalUniqueItems}
+    <Button
+      {...btnProps}
+      className="cart-toggle"
+      aria-label="Button cart toggle"
```
hellofanny, line 15: Do you think we can use another approach to show the quantity? Instead of using a `pseudo-element`, we could use `<span>`  for that. Or maybe you can just pass the quantity on the `aria-label`. Ex: `aria-label="Cart with 2 items"`

renatamottam (commented):
Good job! I left a few points for improvement :)

src/pages/pattern-library.tsx
danzanzini: I got it. My point is regarding using a real context in the pattern library. If it's not too much work, mocking it would be better.

src/components/cart/CartToggle/CartToggle.tsx
ArthurTriis1: I will pass in aria-label!

src/components/cart/CartToggle/cart-toggle.scss
hellofanny: Just use 1px here! 

How about we follow this approach?
If it is less than 4px, just use it.

src/components/cart/CartToggle/cart-toggle.scss
```diff
@@ -0,0 +1,33 @@
+.cart-toggle[data-store-button] {
+  position: relative;
+  background-color: transparent;
+  border-width: 0;
+  cursor: pointer;
+
+  svg {
+    width: var(--space-5);
+    height: var(--space-5);
+  }
```
renatamottam, line 10: You can use `<ShoppingCart size={32} />` instead 🙃

ArthurTriis1: I had done that, but I preferred to leave everything on CSS. Is it better to put in js?

src/components/cart/CartToggle/cart-toggle.scss
```diff
@@ -0,0 +1,28 @@
...
+    padding: 1px var(--space-0);
+    color: var(--color-text-negative);
+    font-weight: var(--text-weight-bold);
+    font-size: var(--text-size-0);
+    background: var(--bg-secondary-default);
+    border-radius: var(--border-radius-pill);
+    content: attr(data-items);
+  }
+
+  img {
```
hellofanny, line 24: Not sure, but do we have img on this component? 👀

hellofanny (approved):
Well done!! 🚀

src/components/cart/CartToggle/cart-toggle.scss
ArthurTriis1: We haven't! I forgot to remove this.

src/components/cart/CartToggle/cart-toggle.scss
```diff
@@ -0,0 +1,23 @@
...
+    align-items: center;
+    justify-content: center;
+    min-width: var(--space-4);
+    padding: 1px var(--space-0);
+    color: var(--color-text-negative);
+    font-weight: var(--text-weight-bold);
+    font-size: var(--text-size-0);
+    background: var(--bg-secondary-default);
+    border-radius: var(--border-radius-pill);
+    content: attr(data-items);
```
eduardoformiga, line 21: Woww, nicee

PR #163, opened 2021-12-21
feat: SkuSelector component

lucasfp13:

This PR intends to add the `SkuSelector` component (styles to be added in a future PR) to be rendered at the Product's details page.

## How it works? 

This component is using the FastStore's `RadioGroup` component and its child components. It is meant to be used as a list of available (or not) variations of a product.

Screen Shot 2021-12-21 at 20 37 24

Screen Shot 2021-12-21 at 20 38 10

### DIFF
--- a/src/components/ui/SkuSelector/SkuSelector.tsx
+++ b/src/components/ui/SkuSelector/SkuSelector.tsx
@@ -22,11 +22,15 @@ function Option({
       <RadioOption
         label={label}
         value={label}
+        style={{ marginLeft: 60 }} // TODO: Remove it. Added just to skip "Tap Target" error (Lighthouse)
         disabled={disabled}
         checked={label === selectedOption}
         {...otherProps}
       >
-        {variant === 'size' && <span>{option.label}</span>}
+        {variant === 'size' && (
+          // TODO: Remove this inline style. Added just to skip "Tap Target" error (Lighthouse)
+          <span style={{ padding: 48 }}>{option.label}</span>
+        )}
         {variant === 'color' && (
           <div style={{ backgroundColor: value, height: 40, width: 40 }} />
         )}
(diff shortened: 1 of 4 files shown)


eduardoformiga (approved):
You nail it! I'm eager to see everything stylized!

src/components/ui/SkuSelector/SkuSelector.tsx
```diff
@@ -0,0 +1,169 @@
...
+   * Section label for the SKU selector.
+   */
+  label?: string
+  /**
+   * Function to be triggered when SKU option change.
+   */
+  onChange?: ChangeEventHandler<HTMLInputElement>
+}
+
+const SkuSelector = ({
```
tlgimenes, line 128: Please use function instead of cost so we don't get lost on the React tree debugger latter on

src/components/ui/SkuSelector/SkuSelector.tsx
```diff
@@ -0,0 +1,169 @@
...
+        selectedValue={selectedSku}
+        name={`sku-selector-${variant}`}
+        onChange={(e) => {
+          onChange?.(e)
+          setSelectedSku(e.currentTarget.value)
+        }}
+      >
+        {options.map((option, index) => {
+          return (
+            <Option
```
tlgimenes, line 155: nit: I prefer personally to not have this component, but rather have the if's etc in a ternary clause in here. I think it's easier to reason about. Something like:
```suggestion
            <RadioOption>{variant === "image" ? <Image/> : <><span> .... </>}</RadioOption>
```

but this is just a personal preference. If  you want you can keep like this :)

tlgimenes (approved):
Overall, LGTM after removing the TODOs

src/components/ui/SkuSelector/SkuSelector.tsx
lucasfp13: @tlgimenes it's a good point, but this probably would make the render body very inflated, as I would still need to do the same type narrowing and we have more than two conditions up there, but I'm looking at a more optimized way of this without having the `<Option />`

saranicoly (approved):

PR #164, opened 2021-12-21
Feat: Creates BuyButton Component

renatamottam:
Creates the [BuyButton](https://vtex-dev.atlassian.net/browse/FSSS-89) component using the FastStore button as base.
This component should only be used on the PDP.

### DIFF
--- a/src/components/ui/BuyButton/BuyButton.tsx
+++ b/src/components/ui/BuyButton/BuyButton.tsx
@@ -1,0 +1,17 @@
+import React from 'react'
+import { Button as UIButton } from '@faststore/ui'
+import type { ButtonProps } from '@faststore/ui'
+import { ShoppingCart } from 'phosphor-react'
+
+type Props = ButtonProps
+
+function BuyButton({ children, ...props }: Props) {
+  return (
+    <UIButton {...props}>
+      <ShoppingCart />
+      {children}
+    </UIButton>
+  )
+}
+
+export default BuyButton
--- a/src/components/ui/BuyButton/BuyButton.tsx
+++ b/src/components/ui/BuyButton/BuyButton.tsx
@@ -1,14 +1,14 @@
 import React from 'react'
 import { Button as UIButton } from '@faststore/ui'
 import type { ButtonProps } from '@faststore/ui'
-import { ShoppingCart } from 'phosphor-react'
+import { ShoppingCart as ShoppingCartIcon } from 'phosphor-react'
 
 type Props = ButtonProps
 
 function BuyButton({ children, ...props }: Props) {
   return (
     <UIButton {...props}>
-      <ShoppingCart />
+      <ShoppingCartIcon />
       {children}
     </UIButton>
   )
--- a/src/pages/pattern-library.tsx
+++ b/src/pages/pattern-library.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 import { GatsbySeo } from 'gatsby-plugin-next-seo'
 import { Button as UIButton, Badge as UIBadge } from '@faststore/ui'
+import BuyButton from 'src/components/ui/BuyButton'
 import { BellRinging } from 'phosphor-react'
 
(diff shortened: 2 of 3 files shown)


hellofanny (approved):

src/components/ui/BuyButton/BuyButton.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { Button as UIButton } from '@faststore/ui'
+import type { ButtonProps } from '@faststore/ui'
+import { ShoppingCart } from 'phosphor-react'
```
danzanzini, line 4: It occurs to me that we may need to have a unique source to retrieve these icons. It has two main reasons:

1) If we decide to change the icon, we will need to change it everywhere it's used. The `ShoppingCart` is already being used at the [CartButton component](https://github.com/vtex-sites/base.store/pull/159), for example.

2) `ShoppingCart` is misleading. It would be good to have some indication that it's an Icon.

I don't think this is immediately necessary. We may tackle it later if we decide to do so. What do you think?

danzanzini (approved):
Looks good to me 🎉. Just left one comment, but it is not a blocker. 

Did you notice that the bot didn't make the comment with the lighthouse results? Is it because it's being merged to `main-styled`?

src/components/ui/BuyButton/BuyButton.tsx
hellofanny: hmm! I think it might be a good idea! We can name it with `Icon` suffix it will be more clear ex: `ShoppingCartIcon`

ArthurTriis1 (approved):

src/components/ui/BuyButton/BuyButton.tsx
```diff
@@ -0,0 +1,17 @@
...
+function BuyButton({ children, ...props }: Props) {
+  return (
+    <UIButton {...props}>
+      <ShoppingCart />
+      {children}
+    </UIButton>
+  )
+}
+
+export default BuyButton
```
eduardoformiga, line 17: Shouldn't we create a generic IconButton here and pass the icon as prop?

```suggestion
import type { ReactNode } from 'react'
import React from 'react'
import { Button as UIButton } from '@faststore/ui'
import type { ButtonProps } from '@faststore/ui'

type Props = ButtonProps & {
  icon: ReactNode
}

function IconButton({ children, icon, ...props }: Props) {
  return (
    <UIButton {...props}>
      {icon}
      {children}
    </UIButton>
  )
}

export default IconButton
```

src/pages/pattern-library.tsx
```diff
@@ -63,7 +64,7 @@ function Page() {
             <h2 className="title-subsection">Buttons</h2>
             <ul className="list-horizontal">
               <li>
-                <UIButton>Call to Action</UIButton>
+                <BuyButton>Buy Now</BuyButton>
```
eduardoformiga, line 67: So we can use as
```suggestion
                <IconButton icon={<ShoppingCart />}>Buy Now</IconButton>
```

eduardoformiga (commented):
Should we create the `BuyButton` or a generic `IconButton`? 
If BuyButton I'm good with the implementation (miss the style).

src/components/ui/BuyButton/BuyButton.tsx
hellofanny: As far as I remember, this BuyButton should be a special button. So I think this is the reason why we aren't using a generic `IconButton`.

src/components/ui/BuyButton/BuyButton.tsx
renatamottam: well noticed, dan. I will fix it right away

src/components/ui/BuyButton/BuyButton.tsx
renatamottam: Yep, this is a button that should only be used on the PDP for the buy action.

PR #175, opened 2021-12-23
Feat: Style Tiles component

eduardoformiga:
- [x] Style the `Tiles` component. 
- [x] Handle the 2 tiles case, so it was necessary to change the component a little.
- [x] Add the component to the `pattern-library`.
- [x] Integrate products in the `Homepage`.
- [x] Integrate `ProductCard` + `Tiles`.
- [x] Integrate `Tiles` in the `Homepage`. 

follow up #141 

* on hold until we can test with internal content, like a `ProductCard`.

## How it works? 

|2 Tile|3 Tile|4 Tile|
|-|-|-|
|tiles 2 items|tiles3|tiles4|
|tiles2|tiles3|tiles4|

## How to test it?
You can see in the root homepage, and also the structure in the `/pattern-library` page

https://preview-175--base.preview.vtex.app/pattern-library/


src/pages/index.tsx
```diff
@@ -19,6 +23,23 @@ function Page(props: Props) {
 
   const title = site?.siteMetadata?.title ?? ''
   const siteUrl = `https://${host}${pathname}`
+  const perPage = 10
+  const productList = useProductsQuery({
+    first: perPage,
+    after: (perPage * 1).toString(),
+    sort: 'score_desc',
+    term: '',
+    selectedFacets: [{ key: 'category-1', value: 'office' }],
```
igorbrasileiro, line 32: Are these selected facets would be intended to be hardcoded?

igorbrasileiro (approved):

src/pages/index.tsx
eduardoformiga: At this moment in an alpha version, this query is to get a real example about how to get products on the homepage.
As we define the products to be shown we can change the query.

src/pages/index.tsx
```diff
@@ -19,6 +23,23 @@ function Page(props: Props) {
...
+  })
+
+  const products = useMemo(
+    () => productList?.edges.map((edge) => edge.node),
+    [productList]
+  )
+
+  if (products == null) {
+    return null
```
filipewl, line 41: Will this make the whole component not render when no products are returned? It seems too drastic if that's the case. Can't we scope this to where `products` is used (in `ProductTiles` below) so it only affects that area of the page? Something like `products && <ProductTiles products={products} />`?

filipewl (approved):
Nice!

Screen Shot 2021-12-30 at 16 08 41

src/pages/index.tsx
eduardoformiga: You are correct! It looks so drastic hahaha. I'm gonna change it. Thanks!

renatamottam (approved):
LATM! (Looks Awesome To Me)

src/components/ui/Tiles/tiles.scss
```diff
@@ -0,0 +1,115 @@
...
+
+[data-store-tiles="expanded-first"] > [data-store-tile]:first-child,
+[data-store-tiles="expanded-first-two"] > [data-store-tile]:first-child,
+[data-store-tiles="expanded-first-two"] > [data-store-tile]:nth-child(2) {
+  grid-column: span 2;
+  min-width: 12rem;
+  height: fit-content;
+
+  [data-store-card-image] {
+    height: calc((100vw - 2rem) * .65);
```
hellofanny, line 89: 👀 How does it work? Is there any value that we can use as variable?

renatamottam: hahahahah it's a fix for now, but we should start using aspect ratio on those images.

hellofanny (approved):
LGTM! (Looks Gorgeous To Me) 🤩

src/components/ui/Tiles/tiles.scss
eduardoformiga: It's a magic made by @renatamottam 🪄  hahahah

eduardoformiga:
Unfortunately, the CI LTTM (Looks Terrify To Me) hahahaha

PR #153, opened 2021-12-17
Chore: Adds include media lib

hellofanny:
## What's the purpose of this pull request?

- Adds [include-media](https://eduardoboucas.github.io/include-media/) to help us to style on breakpoints
- Fixes media rules
- Organizes files
 
## How it works? 
<em>Tell us the role of the new feature, or component, in its context.</em>

## How to test it?
<em>Describe the steps with bullet points. Is there any external reference, link, or example?</em> 

## References
<em>Spread the knowledge: is this any content you used to create this PR that is worth sharing?</em>  

<em>Extra tip: add references to related issues or mention people important to this PR may be good for the documentation and reviewing process</em>

PR #176, opened 2021-12-27
Feat: Style Hero component

ArthurTriis1:
This PR styles the Hero component

## How it works? 
Add a stylesheet for hero and implement a style like the [Figma Prototype](https://www.figma.com/file/wS0Kek1aSzIRl20jdtnSsL/Alpha-Version?node-id=0%3A1)

Screen Shot 2021-12-27 at 12 06 44

Screen Shot 2021-12-27 at 12 06 34

### DIFF
--- a/src/components/ui/Hero/hero.scss
+++ b/src/components/ui/Hero/hero.scss
@@ -38,6 +38,7 @@
     }
 
     [data-store-hero-link] {
+      // TODO: Repassar estilização para Link Button
       margin-top: var(--space-6);
 
       a {
--- a/src/components/ui/Hero/hero.scss
+++ b/src/components/ui/Hero/hero.scss
@@ -1,0 +1,93 @@
+$media: 1280px;
+
+[data-store-hero] {
+  display: flex;
+  width: 100%;
+  background-color: var(--bg-secondary-default);
+
+  @media screen and (max-width: $media) {
+    flex-direction: column-reverse;
+  }
+
+  [data-store-hero-content] {
+    width: 100%;
+    padding: 92px var(--space-9) 92px 105px;
+    color: var(--color-text-negative);
+
+    @media screen and (max-width: $media) {
+      padding: var(--space-6) var(--space-4);
+    }
+
+    .banner-info-content {
+      display: flex;
+      flex-direction: column;
+      justify-content: space-between;
+      max-width: 470px;
+      height: 100%;
+      margin-left: auto;
+
+      @media screen and (max-width: $media) {
+        max-width: 100%;
+      }
+    }
+
+    h1 {
+      font-weight: var(--text-weight-black);
+      font-size: var(--space-8);
+      line-height: var(--space-9);
+
+      @media screen and (max-width: $media) {
+        font-size: var(--space-5);
+        line-height: var(--space-6);
+      }
+    }
+
+    p {
+      margin-top: var(--space-4);
+      font-weight: var(--text-weight-regular);
+      font-size: 20px;
+      line-height: var(--space-5);
+
+      @media screen and (max-width: $media) {
+        font-size: var(--space-3);
+        line-height: var(--space-4);
+      }
+    }
+
+    [data-store-hero-link] {
+      margin-top: var(--space-6);
+
+      a {
+        display: flex;
+        align-items: center;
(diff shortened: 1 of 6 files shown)


src/components/ui/Hero/hero.scss
```diff
@@ -0,0 +1,93 @@
+$media: 1280px;
```
filipewl, line 1: I'm not sure how we're handling media queries but I remember seeing a merged PR related to it. Maybe there's more information in it on how to use them, with predefined Sass variables and stuff. Other people will probably know better.

src/components/ui/Hero/hero.scss
```diff
@@ -0,0 +1,93 @@
...
+
+[data-store-hero] {
+  display: flex;
+  width: 100%;
+  background-color: var(--bg-secondary-default);
+
+  @media screen and (max-width: $media) {
```
filipewl, line 8: Not sure we have some guidelines about this or if it is a big deal, but it looks like we're starting with desktop styles and using media queries to style for mobile. I'd vote for a [mobile-first approach](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Responsive/Mobile_first) instead.

ArthurTriis1: You're right, i'll put in mobile first

src/components/ui/Hero/hero.scss
filipewl: I also wonder if we're off by 1 pixel here. I'm getting the mobile version when my browser is at 1280px (which is half of my total screen width), where at this point it's usually considered fine for desktop.

Screen Shot 2021-12-27 at 12 59 36

src/components/ui/Hero/hero.scss
```diff
@@ -0,0 +1,93 @@
...
+    flex-direction: column-reverse;
+  }
+
+  [data-store-hero-content] {
+    width: 100%;
+    padding: 92px var(--space-9) 92px 105px;
```
filipewl, line 14: I wonder if it's fine to have the `padding-left` to be 0 instead of 105px. It seems to decrease the width for the left-side content in favor of more space for the image (which seems proportionally larger in Figma).

5lpiun3djs

src/components/ui/Hero/hero.scss
```diff
@@ -0,0 +1,93 @@
...
+        font-weight: var(--text-weight-bold);
+        font-size: var(--space-3);
+        text-decoration: none;
+        background: var(--bg-neutral-lightest);
+        border-radius: var(--border-radius-default);
+
+        @media screen and (max-width: $media) {
+          padding: var(--space-1) var(--space-4);
+        }
+      }
```
filipewl, line 76: Looking good! I'm imagining in the future we'll want to extract this somewhere else since this link/button style will be more common in other components (not sure).

filipewl (approved):
Looking good! 👏

src/components/ui/Hero/hero.scss
ArthurTriis1: Is a good idea!

src/components/ui/Hero/hero.scss
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  [data-store-hero-image] {
+    width: 100%;
+
+    [data-gatsby-image-wrapper] {
+      width: 100%;
+      height: 300px;
+      min-height: 0;
```
renatamottam, line 87: Why this `min-height` is needed?

src/components/ui/Hero/hero.scss
renatamottam: True, you can see this PR here: https://github.com/vtex-sites/base.store/pull/153

renatamottam (commented):
Hey Arthur, good job! I left a few suggestions for improvement :)

saranicoly (approved):
LGTM! 👏🏾👏🏾👏🏾

src/components/ui/Hero/hero.scss
renatamottam: I think we should remove this for now and add a `todo` to add the `LinkButton` after it's ready

renatamottam (approved):
Looks great! 👏

PR #179, opened 2021-12-27
Feat: Style Links

renatamottam:
Styles our three types of links:
- Link (default)
- Link Inline
- Link Display
- Link Footer
- Link Inverse (to use on dark backgrounds)

## How it works? 
|Type|Image|
|-|-|
|Default|image|
|Inline|image|
|Display|image|
|Footer|image|
|Inverse|image|

## How to test it?
You can see it working on the `/pattern-library` page


ArthurTriis1 (approved):

src/components/ui/Link/link.scss
```diff
@@ -0,0 +1,40 @@
+@import "../../../styles/vendors/include-media.scss";
+
+[data-store-link] {
```
hellofanny, line 3: Just to confirm, we are applying this style for all the links, right?

src/components/ui/Link/link.scss
```diff
@@ -0,0 +1,40 @@
...
+
+// Links ////////////////////////////
+
+.link {
+  color: var(--color-link);
+  line-height: 1.5;
+}
+
+.link-display {
+  color: var(--color-secondary-4);
```
hellofanny, line 20: Does make sense to create a semantic variable for this color?

src/components/ui/Link/link.scss
```diff
@@ -0,0 +1,40 @@
...
+  line-height: 1.5;
+}
+
+.link-inline {
```
hellofanny, line 24: hmm, here I would suggest or follow BEM pattern to name these classes or create `data-atribule` with variant to style it. What do you think?

hellofanny (changes requested):
Nicee @renatamottam! I've left just a comment 👍  Lemme know what you think and If I can help with that!

israelswf (approved):
Looking and working nicely to me ✨

src/components/ui/Link/link.scss
```diff
@@ -0,0 +1,40 @@
...
+  color: var(--color-link-inverse);
+  &:visited { color: var(--color-link-inverse); }
+}
+
+.link {
+  color: var(--color-link);
+  line-height: 1.5;
+}
+
+.link-display {
```
hellofanny, line 24: Hmm.. if we are going to follow BEM syntax, are we considering these following classes elements or modifies? 
`link__display` or `link--display`?

hellofanny: Or maybe we want to pass all this classes as variants?

hellofanny (changes requested):
hey, I left a comment regarding the variants naming, can you take a look please?

lucasfp13 (approved):

eduardoformiga (approved):

hellofanny (approved):

PR #181, opened 2021-12-28
feat: Add banner text

ArthurTriis1:
This PR styles the BannerText component

## How it works? 
Add a stylesheet for banner text and implement a style like the [Figma Prototype](https://www.figma.com/file/wS0Kek1aSzIRl20jdtnSsL/Alpha-Version?node-id=0%3A1)

Screen Shot 2021-12-28 at 11 08 27

Screen Shot 2021-12-28 at 11 08 14

### DIFF
--- a/src/components/sections/BannerText/banner-text.scss
+++ b/src/components/sections/BannerText/banner-text.scss
@@ -1,0 +1,47 @@
+@import "../../../styles/vendors/include-media.scss";
+
+[data-store-banner] {
+  padding: var(--space-6) var(--space-5);
+  text-align: center;
+  background-color: var(--color-secondary-4);
+
+  @include media(">=notebook") {
+    padding: var(--space-9) 273px;
+  }
+
+  [data-store-banner-content] {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
+
+  h2, p {
+    display: inline;
+    margin-left: var(--space-1);
+    color: var(--color-text-negative);
+
+    @include media(">=notebook") {
+      display: block;
+    }
+  }
+
+  a {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 11.25rem;
+    margin-top: var(--space-5);
+    padding: var(--space-1) var(--space-4);
+    color: var(--color-secondary-4);
+    font-weight: var(--text-weight-bold);
+    font-size: var(--space-3);
+    text-decoration: none;
+    background: var(--bg-neutral-lightest);
+    border-radius: var(--border-radius-default);
+
+    @include media(">=notebook") {
+      margin-top: var(--space-7);
+      padding: var(--space-2) var(--space-4);
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


saranicoly (approved):
LGTM! 😊

renatamottam (approved):
Awesome!
I would just propose to create the other banner color variations, but it's not a block.

src/components/sections/BannerText/banner-text.scss
```diff
@@ -0,0 +1,47 @@
...
+    padding: var(--space-9) 273px;
+  }
+
+  [data-store-banner-content] {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
+
+  h2, p {
```
filipewl, line 18: The paragraph having the same style as the heading is a bit of a red flag for me. 

[On Figma](https://www.figma.com/file/cAXtiQGObWxamFVVhviQVM/Handoff-Homepage-v0.2?node-id=927%3A74363) there's a variation where the paragraph has a smaller font (16px vs the heading's 20px), but that's not our use case here.
Screen Shot 2021-12-28 at 11 51 19

Could this be an indication that we should pass the `BatterText`s `title` as `Receive our news and promotions in advance. Enjoy and get 10% off your first purchase.` and do not pass the `caption`? This way we don't have to style the paragraph the same way as the heading.

Something like:
```jsx
/* Skip the `caption` prop */
<BannerText
  title="Receive our news and promotions in advance. Enjoy and get 10% off your first purchase."
  actionPath="/"
  actionLabel="Call to action"
/>
```
instead of the current:
https://github.com/vtex-sites/base.store/blob/ca403b3bad315be6dcd363c161e1bb7b3d223043/src/pages/index.tsx#L63-L68

filipewl (approved):
I always end up clicking on "Comment" instead of "Approve" because the button is green, lol.

src/components/sections/BannerText/banner-text.scss
ArthurTriis1: In handoff documentation they have differences between title and caption, but in the Alpha version the haven't, so i thought i should do this and change in next version
Screen Shot 2021-12-28 at 12 07 02

Screen Shot 2021-12-28 at 12 07 13

src/components/sections/BannerText/banner-text.scss
```diff
@@ -0,0 +1,47 @@
+@import "../../../styles/vendors/include-media.scss";
+
+[data-store-banner] {
```
hellofanny, line 3: Not sure! But I think we want to scope this only for the `BannerText`? If so we need to add `banner-text` class to  scope it.

src/components/sections/BannerText/banner-text.scss
```diff
@@ -0,0 +1,47 @@
...
+  padding: var(--space-6) var(--space-5);
+  text-align: center;
+  background-color: var(--color-secondary-4);
+
+  @include media(">=notebook") {
+    padding: var(--space-9) 273px;
```
hellofanny, line 9: oops. Can you please convert to rem. (Use `17.06rem` instead o `273px`)

PR #97, opened 2021-11-05
feat:add search event

israelswf:
### Closed PR
This PR was replaced by #182 due to significant refactoring made in the files related to search.

PR #182, opened 2021-12-28
feat: Add search event to Analytics

israelswf:
This PR intends to add the `search` event to the analytics events layer. 

Ps. This PR is replacing the #97.

## How it works? 
Every time a term is searched in the `SearchInput` component, the `sendAnalyticsEvent` is trigged and the `search` event is added to the dataLayer.
## How to test it?
Tests in `analytics.test.js` file will simulate the steps of a search scenario. Just run `yarn test`. 

## References
GA4 reference about [Search event](https://developers.google.com/analytics/devguides/collection/ga4/reference/events#search)

### DIFF
--- a/cypress/integration/analytics.test.js
+++ b/cypress/integration/analytics.test.js
@@ -9,6 +9,11 @@ import { cypress } from '../../store.config'
 
 const { pages } = cypress
 
+beforeEach(() => {
+  cy.clearIDB()
+  cy.log('IDB Cleared')
+})
+
 const dataLayerHasEvent = (eventName) => {
   return cy.window().then((window) => {
     const allEvents = window.dataLayer.map((evt) => evt.name)
@@ -31,10 +36,6 @@ const eventDataHasCurrencyProperty = () => {
 }
 
 describe('add_to_cart event', () => {
-  beforeEach(() => {
-    cy.clearIDB()
-  })
-
   const testAddToCartEvent = (skuId) => {
     cy.window().then((window) => {
       const { dataLayer } = window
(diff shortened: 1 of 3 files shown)


lucasfp13 (approved):
LGTM!

src/components/common/SearchInput/SearchInput.tsx
```diff
@@ -16,6 +21,11 @@ const doSearch = async (term: string) => {
     })
   )
 
+  sendAnalyticsEvent<SearchEvent>({
+    name: 'search',
+    params: { search_term: term },
```
heitorado, line 26: Just out of curiosity, is this the only attribute required/supported for the `search` event?

cypress/integration/analytics.test.js
```diff
@@ -190,3 +190,21 @@ describe('view_item_list event', () => {
     })
   })
 })
+
+describe('search event', () => {
+  beforeEach(() => {
+    cy.clearIDB()
+  })
```
heitorado, line 197: I would suggest moving this to the global scope, just as I did at https://github.com/vtex-sites/base.store/pull/203/commits/d68507d9f9426ac65987dc211eada8900327869c. Just to avoid merging conflicts and/or duplicate code when both get merged.

I think it's repetitive/pointless if we repeat the `cy.clearIDB()` inside each and every `describe` statement. Since we do not want to share state between tests, better move it to the global scope already and make this IDB clearance a default for every new test. This might even make this less error-prone, e.g. where someone forgets to clear the IDB before writing a new tests. 

What do you think?

heitorado (approved):
Looks good! Left some comments but just minor stuff.

cypress/integration/analytics.test.js
israelswf: Makes total sense, Heitor. Thanks.

src/components/common/SearchInput/SearchInput.tsx
israelswf: Apparently it is, Heitor. https://developers.google.com/analytics/devguides/collection/ga4/reference/events#search

PR #189, opened 2021-12-29
style: Add sigin link styles

ArthurTriis1:
This PR styles the SignInLink component

## How it works? 
Add a stylesheet for SignInLink and implement a style like the [Figma Prototype](https://www.figma.com/file/wS0Kek1aSzIRl20jdtnSsL/Alpha-Version?node-id=0%3A1)

Screen Shot 2021-12-29 at 18 03 03

### DIFF
--- a/src/components/ui/Button/LinkButton.tsx
+++ b/src/components/ui/Button/LinkButton.tsx
@@ -1,0 +1,44 @@
+import type { AnchorHTMLAttributes, ReactNode } from 'react'
+import React from 'react'
+import { Icon as UIIcon, Link } from '@faststore/ui'
+
+import './buttons.scss'
+
+type Variant = 'primary' | 'secondary'
+type IconPosition = 'left' | 'right'
+
+type Props = {
+  variant?: Variant
+  inverse?: boolean
+  icon?: ReactNode
+  iconPosition?: IconPosition
+  disabled?: boolean
+} & AnchorHTMLAttributes<HTMLAnchorElement>
+
+function LinkButton({
+  variant,
+  inverse,
+  icon,
+  iconPosition,
+  children,
+  disabled = false,
+  className,
+  ...otherProps
+}: Props) {
+  return (
+    <Link
+      data-store-button
+      className={`button ${className}`}
+      data-button-variant={variant}
+      data-button-inverse={inverse}
+      data-button-disabled={disabled}
+      {...otherProps}
+    >
+      {iconPosition === 'left' && <UIIcon component={icon} />}
+      {children}
+      {iconPosition === 'right' && <UIIcon component={icon} />}
+    </Link>
+  )
+}
+
+export default LinkButton
--- a/src/components/ui/Button/LinkButton.tsx
+++ b/src/components/ui/Button/LinkButton.tsx
@@ -1,19 +1,15 @@
-import type { AnchorHTMLAttributes, ReactNode } from 'react'
 import React from 'react'
-import { Icon as UIIcon, Link } from '@faststore/ui'
+import type { LinkProps } from '@faststore/ui'
+import { Icon as UIIcon, Link as UILink } from '@faststore/ui'
+
+import type { UIButtonProps } from './Button'
 
 import './buttons.scss'
 
-type Variant = 'primary' | 'secondary'
-type IconPosition = 'left' | 'right'
-
 type Props = {
-  variant?: Variant
-  inverse?: boolean
-  icon?: ReactNode
-  iconPosition?: IconPosition
   disabled?: boolean
-} & AnchorHTMLAttributes<HTMLAnchorElement>
+} & UIButtonProps &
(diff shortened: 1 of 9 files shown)


src/components/ui/Button/LinkButton.tsx
```diff
@@ -0,0 +1,44 @@
+import type { AnchorHTMLAttributes, ReactNode } from 'react'
+import React from 'react'
+import { Icon as UIIcon, Link } from '@faststore/ui'
```
hellofanny, line 3: Maybe we can also follow this pattern for Link
`Link as UILink`

src/components/ui/Button/LinkButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Variant = 'primary' | 'secondary'
+type IconPosition = 'left' | 'right'
+
+type Props = {
+  variant?: Variant
+  inverse?: boolean
+  icon?: ReactNode
+  iconPosition?: IconPosition
+  disabled?: boolean
+} & AnchorHTMLAttributes<HTMLAnchorElement>
```
hellofanny, line 16: Why aren't we using `LinkProps` from Link component here?

src/components/ui/Button/LinkButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+function LinkButton({
+  variant,
+  inverse,
+  icon,
+  iconPosition,
+  children,
+  disabled = false,
+  className,
```
hellofanny, line 25: Do you mean other classes that this component will recieve? If so, I think it's better to use `classes` instead of `classNames`. What do you think?

src/components/ui/Button/buttons.scss
```diff
@@ -1,23 +1,34 @@
...
     cursor: not-allowed;
+    pointer-events: none;
     svg { color: var(--color-neutral-5); }
   }
 
+  &:hover {
+    text-decoration: none;
+  }
+
+  &:visited {
```
hellofanny, line 28: Maybe we can separate this a bit to make it less confusing.

Does it make sense to add a `button-link` class on this component just to toggle link specif attributes? 🤔

src/components/ui/Button/LinkButton.tsx
ArthurTriis1: i think this component can receive other classes, if a user needs to do a specific style, why not className?  Wouldn't it be better to stick with naming pattern?

hellofanny: hmm, got it! I'm just wondering if it's going to be confusing, but I'm fine with that.

src/components/ui/Button/buttons.scss
```diff
@@ -63,4 +66,39 @@
...
+
+    &:hover, &:focus { background-color: var(--bg-secondary-light); }
+
+    &:active {
+      color: var(--color-text-display);
+      background-color: var(--color-secondary-1);
+    }
+
+    &:disabled, &[data-button-disabled="true"] {
+      color: var(--color-neutral-6);
```
hellofanny, line 81:
```suggestion
      color: var(--color-text-disabled);
```

src/components/ui/Button/buttons.scss
```diff
@@ -63,4 +66,39 @@
...
+      background-color: var(--bg-disabled);
+    }
+
+    &[data-button-inverse] {
+      color: var(--color-text-display);
+      background-color: var(--bg-neutral-lightest);
+
+      &:hover, &:focus { background-color: var(--color-secondary-0); }
```
hellofanny, line 89:
```suggestion
      &:hover, &:focus { background-color: var(--bg-secondary-light); }
```

hellofanny (approved):
Niceee!

src/components/ui/Button/buttons.scss
```diff
@@ -63,4 +66,39 @@
...
+    }
+  }
+}
+
+.link-button[data-store-button] {
+  &:hover {
+    text-decoration: none;
+  }
+
+  &:disabled, &[data-button-disabled="true"] {
```
renatamottam, line 101: Could you leave the `pointer-events: none;` only for `[data-button-disabled="true"]`? That way the `cursor: not-allowed` will work for the `<button>` at least.

src/components/ui/SignInLink/SignInLink.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const SignInLink: React.FC = () => {
+  return (
+    <LinkButton
+      data-button-signin-link
+      href="/"
+      className="title-sub-subsection signin-link"
+      variant="tertiary"
+    >
+      <UserIcon size={24} />
```
renatamottam, line 14:
```suggestion
      <UserIcon size={18} weight="bold" />
```

renatamottam (approved):
Great job, Arthur!

PR #192, opened 2021-12-30
Feat: Style Incentives list

hellofanny:

- Incentives Component (Generic)
- IncentivesHeader (Uses Incentives component to custom and added to homepage)

|Mobile|Desktop|
|-|-|
|mobile-preview|desktop-preview

OBS: For now, we have an accessibility issue with this component because the list is scrollable in mobile, but there is no indicator nor focus on this element. We will keep it like that because this is not the final version of this component.

## How it works? 
<em>Tell us the role of the new feature, or component, in its context.</em>

## How to test it?

## References


renatamottam (approved):
Looks beautiful 😍 I left a suggestion, but it's perfect otherwise

src/components/sections/Incentives/Incentives.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface Props {
+  incentives: Incentive[]
+  hasBackground?: boolean
+}
+
+function Incentives({ incentives, hasBackground }: Props) {
+  return (
+    <div
+      className="incentives / grid-content-full"
+      data-incentives-background={hasBackground}
```
eduardoformiga, line 22: If we pass `hasBackground={false}`, the background is applied yet. Maybe we can check something like:
```suggestion
      data-incentives-background={hasBackground || undefined}
```

after that the `eslint` is gonna suggest to use `nullish-coalescing (??) operator` instead of `logical (||) operator`, but the results are different. So I recommend disabling this rule in the `.eslint`

```
"rules": {
    "@typescript-eslint/prefer-nullish-coalescing": "off"
  }
```

I wonder if instead of a boolean flag we can pass either the `background-color` and styles inline or `variants` (class) with CSS ready. But it's not blocking, it's up to you hahah.

eduardoformiga (changes requested):
Good Job! 👏 
I left one suggestion about the `hasBackground={false}`

src/components/sections/Incentives/incentives.scss
```diff
@@ -0,0 +1,75 @@
+@import "../../../styles/scaffold";
+
+.incentives {
+  [data-store-list] {
+    display: grid;
+    grid-template-columns: repeat(5, 1fr);
```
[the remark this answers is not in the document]
hellofanny: Oh sure! I'll change that 😉

src/components/sections/Incentives/Incentives.tsx
hellofanny: hmm, maybe we can pass a class here. I'll try that! Thanks!

eduardoformiga (approved):
🤩 🤩 🤩

PR #201, opened 2022-01-04
feat: Adds SROnly component

hellofanny:
- "sr-only" means "screen reader only"
- Introducing this technique for hiding content when other approaches are unavailable. We should use this component to give extra context that could be visibly hidden in the interface.
- Aims to deliver better experience for screen readers users.

## How it works? 
- Visually, nothing changes.

https://user-images.githubusercontent.com/3356699/148112592-4223b85b-2051-419d-b3b0-e3e000d43e47.mov

## References
- https://webaim.org/techniques/css/invisiblecontent/

### DIFF
--- a/src/components/product/ProductCard/ProductCard.tsx
+++ b/src/components/product/ProductCard/ProductCard.tsx
@@ -101,7 +101,7 @@ function ProductCard({
               data-value={listPrice}
               variant="listing"
               classes="text-body-small"
-              SRText="Original price:"
+              sr-text="Original price:"
             />
             <Price
               value={spotPrice}
--- a/src/components/product/ProductCard/ProductCard.tsx
+++ b/src/components/product/ProductCard/ProductCard.tsx
@@ -101,6 +101,7 @@ function ProductCard({
               data-value={listPrice}
               variant="listing"
               classes="text-body-small"
+              SRText="Original price:"
             />
             <Price
               value={spotPrice}
@@ -109,6 +110,7 @@ function ProductCard({
               data-value={spotPrice}
               variant="spot"
               classes="text-body"
+              SRText="Sale Price:"
             />
           </div>
         </div>
--- a/src/components/product/ProductCard/ProductCard.tsx
+++ b/src/components/product/ProductCard/ProductCard.tsx
@@ -101,7 +101,7 @@ function ProductCard({
               data-value={listPrice}
               variant="listing"
               classes="text-body-small"
-              sr-text="Original price:"
+              SRText="Original price:"
             />
             <Price
               value={spotPrice}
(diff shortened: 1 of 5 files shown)


src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -101,6 +101,7 @@ function ProductCard({
               data-value={listPrice}
               variant="listing"
               classes="text-body-small"
+              SRText="Original price:"
```
eduardoformiga, line 104: Can we use the `kebab-case` here to follow the pattern?
```suggestion
              sr-text="Original price:"
```

src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -109,6 +110,7 @@ function ProductCard({
               data-value={spotPrice}
               variant="spot"
               classes="text-body"
+              SRText="Sale Price:"
```
eduardoformiga, line 113:
```suggestion
              sr-text="Sale Price:"
```

src/components/ui/Price/Price.tsx
```diff
@@ -2,15 +2,28 @@ import React from 'react'
...
+import SROnly from '../SROnly'
+
 import './price.scss'
 
 type Props = PriceProps & {
-  // other classes
+  /**
+   * Text for the screen readers only
+   */
+  SRText: string
```
eduardoformiga, line 13:
```suggestion
  'sr-text': string
```

src/components/ui/SROnly/sr-only.scss
```diff
@@ -0,0 +1,9 @@
+[data-store-sr-only]:not(:focus):not(:active) {
```
eduardoformiga, line 1: 🪄

eduardoformiga (approved):
LGTM! 👏  left a comment regarding the `kebab-case`.

At first glance, the name `SROnly` looks strange to me.
I wonder if we can use something like `ScreenReaderOnly`, but not sure because it can be very common to use `SROnly` in other places.

hellofanny:
> LGTM! 👏 left a comment regarding the `kebab-case`.
> 
> At first glance, the name `SROnly` looks strange to me. I wonder if we can use something like `ScreenReaderOnly`, but not sure because it can be very common to use `SROnly` in other places.

Oh sure! Updated! 
hmm, It's just because it's common to see `SROnly`, but we can rename it too.

eduardoformiga:
If it's common to see let's keep `SROnly` 😉

eduardoformiga:
Hey @hellofanny, my bad here. 
I double-checked the pattern and we are using `cameCame` instead of `kebab-case`.
sorry for this confusion hahaha

renatamottam (approved):

PR #202, opened 2022-01-04
chore: Adjusts main-styled

eduardoformiga:
This PR adjusts the `main-styled` so that we can merge into the `master` branch afterward.

	
- [x] Remove `pattern-library`.
- [x] Review FIXMEs.
- [x] Review temporary classes (`temp-grid-row`, `temp-offscreen`).
- [x] Text Overflow (Clamp) `ProductCard` (`Tiles` too).
- [x] Replace `Hero` and `TextBanner` with `LinkButton`.
- [x] Adds minimum acceptable CSS to `ProductGrid`.

## References
[FSSS-104](https://vtex-dev.atlassian.net/jira/software/projects/FSSS/boards/632?selectedIssue=FSSS-104)


src/pages/s.tsx
```diff
@@ -56,7 +56,7 @@ function Page(props: Props) {
         Sections: Components imported from '../components/sections' only.
         Do not import or render components from any other folder in here.
       */}
-      <h1 className="temp-offscreen">{title}</h1>
+      <h1>{title}</h1>
```
filipewl, line 59: This is causing the following on PDPs:
CleanShot 2022-01-12 at 20 05 13

src/pages/{StoreProduct.slug}/p.tsx
```diff
@@ -96,7 +96,7 @@ function Page(props: Props) {
         Sections: Components imported from '../components/sections' only.
         Do not import or render components from any other folder in here.
       */}
-      <h1 className="temp-offscreen">{title}</h1>
+      <h1>{title}</h1>
```
filipewl, line 99: This is causing the following on PDPs:
CleanShot 2022-01-12 at 20 05 13

Similar issue on PLPs.

filipewl (approved):
Compared this (https://preview-202--base.preview.vtex.app/) side-by-side with `main-styled` (https://preview-131--base.preview.vtex.app/) and noticed a few minor things:

1. The mini-cart counter seems misaligned now.
2. The links on the footer have less space between each other.

1|2
-|-
CleanShot 2022-01-12 at 20 00 02|CleanShot 2022-01-12 at 20 03 17

src/pages/{StoreProduct.slug}/p.tsx
hellofanny: That's true! I was wondering what would be the best approach for this. It's related to the SEO right?
Maybe we can create a class just for that, maybe we can use the `SROnly` component (because it hides the text from the page, but it's not for this purpose). 🤔

src/components/product/ProductCard/product-card.scss
```diff
@@ -49,11 +49,23 @@
   }
 
   [data-card-content] a[href] {
+    display: -webkit-box;
```
hellofanny, line 52: I don't think we need to apply those prefixes, it's added automatically 🤔  `-webkit`, `-moz` ...  Can you double check if it's working?

src/pages/{StoreProduct.slug}/p.tsx
eduardoformiga: That's true. I have just removed the temporary class so far. 
Is this title required? I noticed that on the `Homepage` we don't have it.
And when we add the `Hero` on `PLP` we are going to have an `h1` tag.
Thoughts?

eduardoformiga:
> Compared this (https://preview-202--base.preview.vtex.app/) side-by-side with `main-styled` (https://preview-131--base.preview.vtex.app/) and noticed a few minor things:
> 
> 1. The mini-cart counter seems misaligned now.
> 2. The links on the footer have less space between each other.
> 


Hmm, that's true, @filipewl.
And I don't know why but looks like [the CSS from the pattern library](https://github.com/vtex-sites/base.store/blob/main-styled/src/styles/pattern-library.scss#L81-L86) is affecting other CSS, even on different pages.
I'll fix these styles, but let's keep the eyes open about that 👀

src/components/product/ProductCard/product-card.scss
eduardoformiga: This value is required for the `line-clamp`. 
When I was testing it, I notice that it only works with the `-webkit` prefix.
I think the `autoprefixer` is not being able to add the prefixes for `box` value. Maybe because is an old draft of the new flexbox. But there is no solution using flexbox, though.

There is this warning on [autoprefixer github](https://github.com/postcss/autoprefixer#warnings). So I decided to keep it so far. 
I tested using Chrome, Sarafi, and Firefox.

Any suggestion?

hellofanny: Oh! Got it! So let's keep it :)

src/pages/{StoreProduct.slug}/p.tsx
hellofanny: I think it was added for SEO purposes.
Hmm. Maybe we could change this title on PDP for an `h1` tag?
image

Not sure; maybe the safest approach now could be: create a new class just for that and keep it.

lucasfp13 (approved):
Great job!

src/pages/{StoreProduct.slug}/p.tsx
eduardoformiga: We remove this `h1` tag on PLP since we already have it on the `Hero`. 
On the PDP I agree that we can turn this `h2` into `h1` since we don't have another more "important" option.

PR #204, opened 2022-01-05
Chore: Simplify images

tlgimenes:
This PR removes that `image.conf` file where all images config were centralized. Also, this removes the anti DDOS security we had on our image service. Now, the server is opened to the wild.

## How it works? 
By removing the image config file and placing the config in the component declaration, we hope to make it more simple to handle images on gatsby. This, however, comes at the price of opening our image service to the world, making it prone to DDOS attacks. This, however, I see as a minor security issue, since Next.JS apps are prone to this attack and they are doing just fine on the internet so far.

## How to test it?
Make sure that everything works just fine. Nothing should have changed.


package.json
```diff
@@ -29,7 +29,7 @@
     "@faststore/sdk": "^1.3.28",
     "@faststore/ui": "^1.3.28",
     "@vtex/gatsby-plugin-nginx": "^1.3.22",
-    "@vtex/gatsby-plugin-thumbor": "^1.3.22",
+    "@vtex/gatsby-plugin-thumbor": "https://pkg.csb.dev/vtex/faststore/commit/47ec99e8/@vtex/gatsby-plugin-thumbor",
```
heitorado, line 32: Why are we not using the version number here anymore? What does this mean?

src/components/ui/Image/Image.tsx
```diff
@@ -1,23 +1,55 @@
...
 import React, { useMemo } from 'react'
-import imagesConf from 'src/images/config'
+import type { ThumborImageOptions } from '@vtex/gatsby-plugin-thumbor'
 import type { GatsbyImageProps } from 'gatsby-plugin-image'
 
-interface Props extends Omit<GatsbyImageProps, 'image'> {
-  variant: string
-  src: string
-}
+interface Props extends Omit<GatsbyImageProps, 'image'>, ThumborImageOptions {}
```
heitorado, line 7: What do you think about making this `Props` interface so it better separates which properties are from `Thumbor` and which are from `GatsbyImage`? This way it might be easier to fiddle with this component later if someone decides to use an image service other than `Thumbor` or if we do not use `GatsbyImage` anymore.

Maybe something like this?
```TypeScript
interface Props extends Omit<GatsbyImageProps, 'image'> {
  thumborOptions: ThumborImageOptions
}
```

And then we could call the `Image` component like:
```jsx
// I don't know exactly which attributes are from GatsbyImage only
<Image baseUrl={...} /*  other GatsbyImage/img attributes here */ thumborOptions={{...}} />
```

src/components/ui/Image/ImageWithArtDirection.tsx
```diff
@@ -1,42 +1,27 @@
...
-  }
-  mobile: {
-    src: string
-    variant: string
-  }
+  desktop: ThumborImageOptions
+  mobile: ThumborImageOptions
 }
 
 function ImageWithArtDirection({ mobile, desktop, ...imgProps }: Props) {
```
heitorado, line 12: Out of curiosity, what is the purpose of this `ImageWithArtDirection` component?

gatsby-config.js
```diff
@@ -17,20 +16,6 @@ const {
...
-  unique(
-    Object.values(variants)
-      .flatMap((variant) =>
-        variant.breakpoints.map((width) => [
-          `${width}x${Math.ceil(width / variant.aspectRatio)}`,
-          `${width}x${Math.floor(width / variant.aspectRatio)}`,
-        ])
-      )
-      .flat()
-  )
```
heitorado, line 32: Out of curiosity: what does this `getSizes` method was doing before, exactly? How/where/by whom these values were used?

heitorado (changes requested):
Looking good, I feel we're on the right path to remove this image complexity. We just need to do some more adjustments to make the parameters more "obvious" and maybe decouple things a bit more so we can easily switch between services and the underlying image component, I think.

package.json
tlgimenes: It's using this PR: https://github.com/vtex/faststore/pull/1098. After merging it in there we can update the version in here.

src/components/ui/Image/ImageWithArtDirection.tsx
tlgimenes: take a look in here: https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-plugin-image/#withartdirection

gatsby-config.js
tlgimenes: It was used by the proxy to avoid DDOS attacks. Now we are prone to this attack because there is no proxy in between the browser and the image server to avoid people resizing images

src/components/ui/Image/Image.tsx
tlgimenes: I'm not sure. This would case using the component a little bit weird, because this `baseURL` should go inside `thumborOptions`, even though this baseURL is used by GatsbyImage hook and not by thumbor itself. Look, an example:
```tsx
// How it's with current implemetation:
<Image
  baseUrl={productImages[0].url}
  sourceWidth={720}
  aspectRatio={1}
  width={720}
  breakpoints={[250, 360, 480, 720]}
  layout="constrained"
  backgroundColor="#f0f0f0"
  options={{
    fitIn: true,
  }}
  alt={productImages[0].alternateName}
  loading="eager"
/>

// How it will be after the suggested changes
<Image
  thumborOptions={{
    baseUrl: productImages[0].url,
    sourceWidth: 720,
    aspectRatio: 1,
    width: 720,
    breakpoints: [250, 360, 480, 720],
    layout: "constrained",
    backgroundColor: "#f0f0f0",
    options: {
      fitIn: true,
    },
  }}
  alt={productImages[0].alternateName}
  loading="eager"
/>
```

src/components/ui/Image/Image.tsx
```diff
@@ -1,23 +1,59 @@
...
+import type { ThumborImageOptions } from '@vtex/gatsby-plugin-thumbor'
 import type { GatsbyImageProps } from 'gatsby-plugin-image'
 
 interface Props extends Omit<GatsbyImageProps, 'image'> {
-  variant: string
-  src: string
+  image: ThumborImageOptions
 }
 
-function Image({ variant, src, ...imgProps }: Props) {
```
icazevedo, line 12: Does Thumbor offer a playground so people can preview and learn what each of these options does? If it does, I think it's worth referencing it here.

icazevedo (approved):
LGTM

src/components/ui/Image/Image.tsx
tlgimenes: These are not thumbor props, but gatsby image ones. The thumbor one are those passed in options, like `fit-in` properties

heitorado (approved):

PR #208, opened 2022-01-06
Chore: Tweaks badge value

eduardoformiga:
This PR aims to round the value from the discount badge.

ps1: We need to double-check whether we should round to the close number, the floor, or the ceil.
ps2: On Hold until we check that we can do that.

|Before|After|
|-|-|
|Screen Shot 2022-01-06 at 13 01 33|Screen Shot 2022-01-06 at 13 02 20|


hellofanny (approved):
LGTM :)

filipewl (approved):
LGTM!

PR #210, opened 2022-01-06
feat: Add AspectRatio component

lucasfp13:

This PR intends to add the `AspectRatio` component.

## References

- [Chakra UI](https://chakra-ui.com/docs/layout/aspect-ratio);
- [Radix UI](https://www.radix-ui.com/docs/primitives/components/aspect-ratio);
- ["Let's learn about aspect ratio in CSS"](https://ishadeed.com/article/css-aspect-ratio/);
- [CSS Tricks](https://css-tricks.com/almanac/properties/a/aspect-ratio/).

### DIFF
--- a/src/components/product/ProductCard/ProductCard.tsx
+++ b/src/components/product/ProductCard/ProductCard.tsx
@@ -4,6 +4,7 @@ import Button from 'src/components/ui/Button'
 import DiscountBadge from 'src/components/ui/DiscountBadge'
 import Price from 'src/components/ui/Price'
 import AspectRatio from 'src/components/ui/AspectRatio'
+import type { AspectRatioProps } from 'src/components/ui/AspectRatio'
 import { Image } from 'src/components/ui/Image'
 import { useBuyButton } from 'src/sdk/cart/useBuyButton'
 import { useFormattedPrice } from 'src/sdk/product/useFormattedPrice'
@@ -26,7 +27,7 @@ interface Props {
   index: number
   variant?: Variant
   showActions?: boolean
-  ratio?: string
+  ratio?: AspectRatioProps['ratio']
 }
 
 function ProductCard({
--- a/src/components/product/ProductCard/ProductCard.tsx
+++ b/src/components/product/ProductCard/ProductCard.tsx
@@ -3,6 +3,7 @@ import React, { useMemo } from 'react'
 import Button from 'src/components/ui/Button'
 import DiscountBadge from 'src/components/ui/DiscountBadge'
 import Price from 'src/components/ui/Price'
+import AspectRatio from 'src/components/ui/AspectRatio'
 import { Image } from 'src/components/ui/Image'
 import { useBuyButton } from 'src/sdk/cart/useBuyButton'
 import { useFormattedPrice } from 'src/sdk/product/useFormattedPrice'
@@ -25,6 +26,7 @@ interface Props {
   index: number
   variant?: Variant
   showActions?: boolean
+  ratio?: string
 }
 
 function ProductCard({
@@ -32,6 +34,7 @@ function ProductCard({
   index,
   variant = 'vertical',
   showActions = true,
+  ratio = '1',
 }: Props) {
   const {
     id,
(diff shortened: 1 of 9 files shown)


src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -75,20 +78,22 @@ function ProductCard({
...
-          breakpoints={[250, 360, 480]}
-          layout="constrained"
-          backgroundColor="#f0f0f0"
-          options={{
-            fitIn: true,
-          }}
-          alt={img.alternateName}
-          sizes="(max-width: 768px) 200px, 320px"
-        />
+        <AspectRatio ratio={ratio}>
```
tlgimenes, line 81: Why do we need this component? Can't we use `aspect-ratio` css property?

lucasfp13: Basically, because the `aspect-ratio` property isn't fully supported yet. If you take a look at [Can I Use](https://caniuse.com/?search=aspect-ratio) results we can see that the property only has ~81% global coverage, which isn't interesting

src/components/ui/AspectRatio/AspectRatio.tsx
```diff
@@ -0,0 +1,36 @@
...
+   * ID to find this component in testing tools (e.g.: cypress,
+   * testing-library, and jest).
+   */
+  testId?: string
+  /*
+   * The aspect ratio value: Common values are:
+   *
+   * `21:9`, `16:10`, `16:9`, `4:3`, `2:1`
+   */
+  ratio?: string
```
eduardoformiga, line 17: I was wondering if we can use these values as `type Ratio = "21:9" | "16:10" | "16:9" | "4:3" | "2:1" [...]`. But I agree that string keep it more flexible :)

src/components/ui/AspectRatio/AspectRatio.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const AspectRatio = forwardRef<HTMLDivElement, Props>(function AspectRatio(
+  { children, ratio = '1', testId = 'store-aspect-ratio', ...otherProps },
+  ref
+) {
+  // Ensures only one child component
+  const child = Children.only(children)
```
eduardoformiga, line 25: 💯

eduardoformiga (approved):
Awesome! 🤩

saranicoly (approved):
Nice job!

filipewl (commented):
I noticed some images with odd dimensions [on the PLP](https://preview-210--base.preview.vtex.app/office/), is that expected?

CleanShot 2022-01-11 at 14 28 33

src/components/ui/AspectRatio/AspectRatio.tsx
lucasfp13: hmm I think we can add your suggestion right now because we don't need to be flexible here as we are not dealing with custom ratios yet

eduardoformiga (approved):

src/components/ui/AspectRatio/AspectRatio.tsx
```diff
@@ -0,0 +1,63 @@
...
+const AspectRatio = forwardRef<HTMLDivElement, AspectRatioProps>(
+  function AspectRatio(
+    { children, ratio = '1', testId = 'store-aspect-ratio', ...otherProps },
+    ref
+  ) {
+    // Ensures only one child component
+    const child = Children.only(children)
+
+    return (
+      <div data-store-aspect-ratio-wrapper data-variant={ratio} {...otherProps}>
```
hellofanny, line 54: can't this be `data-store-aspect-ratio` and the other one `data-aspect-ratio-content` ?

src/components/ui/AspectRatio/aspect-ratio.scss
```diff
@@ -0,0 +1,109 @@
+[data-store-aspect-ratio-wrapper] {
+  position: relative;
+  display: block;
+  width: 100%;
+  height: 0;
+  overflow: hidden;
+
+  &[data-variant="1"] {
```
hellofanny, line 8: Do you think it's better if we split `Ratio` and `InverseRatio` values here. Or it's okay to keep them listed all together?  
(Ex: adding a comment to separate them) If it's not needed I'm also fine with this list :)

hellofanny (approved):
Well done! 🚀 🚀 🚀

PR #211, opened 2022-01-06
Fix/event triggers and params

icazevedo:
This PR fixes where some events are triggered and also standardizes the parameters sent with the events. They are now equal and the properties match the appropriate values.

## How it works? 
We're using the new types introduced by https://github.com/vtex/faststore/pull/1101 to type the `sendAnalyticsEvent` function and sending the product id as the `item_id`, including the discount and right index values on events, etc.

## How to test it?
Check the dataLayer variable on the console tab and check if the event parameters match each other. 

## References
Depends on https://github.com/vtex/faststore/pull/1101


heitorado (approved):
Looking good!

tlgimenes (approved):

PR #217, opened 2022-01-10
Feat: style load more button

eduardoformiga:
This PR aims to create and style the `Show More ButtonLink` on `plp` page.

** we decide to keep the original LinkButton styles (Padding). So we reverted the `ShowMoreLink` component.

|Before|After|
|-|-|
|fb2f17ab-f573-4487-a3ca-2ebdd084df7f|Screen Shot 2022-01-10 at 21 31 15|

## How to test it

https://preview-217--base.preview.vtex.app/office/


src/components/sections/ProductGallery/ProductGallery.tsx
```diff
@@ -66,18 +67,16 @@ function ProductGallery({ title }: Props) {
       {next !== false && (
         <>
           <GatsbySeo linkTags={[{ rel: 'next', href: next.link }]} />
-          <a
-            data-testid="show-more"
+          <ShowMoreLink
```
hellofanny, line 70: @eduardoformiga actually, I don't think we need to create a new component just for that.
The only issue was because of the style. I talked with renatinha. Let's keep the default version (it doesn't need the extra padding on the left/right sides) 🤙 

 ``<LinkButton .... variant="secondary">
     Load more products
  </LinkButton>``

eduardoformiga: Yeahh, Agree with you.
Let's change to keep it simple :)

lucasfp13 (approved):

src/components/sections/ProductGallery/ProductGallery.tsx
```diff
@@ -66,7 +67,7 @@ function ProductGallery({ title }: Props) {
       {next !== false && (
         <>
           <GatsbySeo linkTags={[{ rel: 'next', href: next.link }]} />
-          <a
+          <LinkButton
```
igorbrasileiro, line 70: IMO, it should be an anchor. Buttons don't handle opening new tabs, and anchor has a better semantic for the scenario since this "button" controls navigation.

igorbrasileiro (approved):

PR #220, opened 2022-01-11
Chore: enhance GraphQL

tlgimenes:
This PR brings a few enhancements to the GraphQL layer. 
The first enhancement is related to testing dynamic queries. Before, there was not really a way for testing dynamic queries like `search`, `allProducts` etc because we forced all queries to be known from the store's code before executing it (through persisted queries). This increased the security (because we only run known queries), however, made a poor DX. This PR makes it possible to query a full text graphql query during development only, so people can use tools like [GraphQL Playground](https://github.com/graphql/graphql-playground) for building their queries and than use it on their code

The second enhancement is `console.logging` masked errors before they are masked, so users can know exactly what error happened during development. For instance, the console output before and after this PR:
```sh
 ERROR

GraphQLError [Object]: Sorry, something went wrong.
    at maskError (webpack://base.store/./src/api/graphql.ts?:42:14)
    at eval (webpack://base.store/./node_modules/@envelop/core/index.mjs?:1084:67)
    at Array.map (<anonymous>)
    at eval (webpack://base.store/./node_modules/@envelop/core/index.mjs?:1084:54)
    at handleStreamOrSingleExecutionResult

```
The error above is masked and the developer doesn't really know what is happening on the backend. However, after the changes:
```sh
 ERROR

NotFoundError: Not Found: kitchen-and-home-appliancesasd
    at Object.collection
(webpack://base.store/./node_modules/@faststore/api/dist/api.esm.js?:798:11)
    at processTicksAndRejections (node:internal/process/task_queues:93:5) {
  locations: [ { line: 2, column: 3 } ],
  path: [ 'collection' ]
}

Executed function "/api/graphql" in 719ms
```

## How it works? 
Most of the changes were done on two files, `graphql.ts` the function, and `server/index.js`, the core server implementation.

## How to test it?
Make sure nothing has really changed and everything works just fine


src/server/index.js
```diff
@@ -1,28 +1,92 @@
+/* eslint-disable react-hooks/rules-of-hooks */
 const {
-  getSchema: storeApiGetSchema,
-  getContextFactory: storeApiGetContextFactory,
-} = require('@faststore/api')
+  envelop,
```
igorbrasileiro, line 3: Why does this file need to be a js one? Can't do something like this, can?

```ts
// src/server/index.ts
...
export {}
```

```js
// /src/server/index.js
const { } = require('./dist/server/index')
```

src/server/index.js
```diff
@@ -1,28 +1,92 @@
...
+const apiContextFactory = getContextFactory(apiOptions)
+
+const isBadRequestError = (err) => {
+  return err.originalError && err.originalError.name === 'BadRequestError'
+}
 
-const getSchema = () => schema
-const getContextFactory = () => contextFactory
+const formatError = (err) => {
+  console.error(err)
```
igorbrasileiro, line 34: Does it make sense to log on production environment? If not, can you remove it if the environment is production?

src/api/graphql.ts
```diff
@@ -1,52 +1,27 @@
...
 
 const parseRequest = (req: GatsbyFunctionRequest) => {
-  const { operationName, variables } =
+  const { operationName, variables, query } =
     req.method === 'POST'
       ? req.body
       : {
           operationName: req.query.operationName,
-          variables: JSON.parse(req.query.variables),
+          variables: JSON.parse(req.query.variables.toString()),
```
victorhmp, line 11: Could `req.query.variables` be `undefined` in any case? If so, just not calling `toString()` would pass the `undefined` on and work fine right? How about 

```
variables: JSON.parse(req.query.variables?.toString()),
```

?

src/server/index.js
```diff
@@ -1,28 +1,92 @@
...
+
+  if (err instanceof GraphQLError) {
+    if (!isBadRequestError(err)) {
+      return new GraphQLError('Sorry, something went wrong.')
+    }
+
+    return err
+  }
```
igorbrasileiro, line 42:
```suggestion
  if (err instanceof GraphQLError && isBadRequestError(err)) {
    return err
  }
```

Make the code branchless as much as possible.

src/api/graphql.ts
```diff
@@ -57,32 +32,24 @@ const handler = async (
...
+      },
+      { req }
+    )
+
+    if (Array.isArray(response.errors)) {
+      // TODO: Return 400 on userError
+      res.status(500)
     }
 
+    res.setHeader('cache-control', 'no-cache, no-store')
```
victorhmp, line 52: Should this header be added in production too?

victorhmp (changes requested):

igorbrasileiro (approved):

src/api/graphql.ts
tlgimenes: Yes, we are not caching graphql data for now

src/server/index.js
tlgimenes: it needs to be js because `gatsby-config.js` uses it

src/server/index.js
tlgimenes: nice!

victorhmp (approved):

src/api/graphql.ts
```diff
@@ -1,52 +1,27 @@
...
 
 const parseRequest = (req: GatsbyFunctionRequest) => {
-  const { operationName, variables } =
+  const { operationName, variables, query } =
     req.method === 'POST'
       ? req.body
       : {
           operationName: req.query.operationName,
-          variables: JSON.parse(req.query.variables),
+          variables: JSON.parse(req.query.variables?.toString()),
```
emersonlaurentino, line 11:
```suggestion
          variables: req.query.variables?,
```

Why do you need to transform to string to parse to JSON?

emersonlaurentino (approved):

src/api/graphql.ts
tlgimenes: I think we don't need it. Nice!

PR #228, opened 2022-01-12
feat: Add Filter component

lucasfp13:

This PR intends to add the `Filter` component.

### DIFF
--- a/gatsby-ssr.js
+++ b/gatsby-ssr.js
@@ -42,6 +42,12 @@ export const onRenderBody = ({ setHeadComponents }) => {
         enablePartytown
       />,
       <Partytown key="party" />,
+      <link
+        key="preconnect"
+        rel="preconnect"
+        crossOrigin
+        href="https://googletagmanager.com"
+      />,
     ])
   } else if (process.env.NODE_ENV === 'development') {
     // eslint-disable-next-line
--- a/src/components/search/Filter/Filter.tsx
+++ b/src/components/search/Filter/Filter.tsx
@@ -14,6 +14,7 @@ import {
   AccordionButton as UIAccordionButton,
   AccordionPanel as UIAccordionPanel,
 } from '@faststore/ui'
+import useWindowDimensions from 'src/hooks/useWindowDimensions'
 import Button from 'src/components/ui/Button'
 import Checkbox from 'src/components/ui/Checkbox'
 
@@ -28,10 +29,21 @@ interface Props {
    * the filter modal or clicks in close button. (mobile only)
    */
   onDismiss?: (event: MouseEvent | KeyboardEvent) => void
+  /**
+   * ID to find this component in testing tools (e.g.: cypress,
+   * testing-library, and jest).
+   */
+  testId?: string
 }
 
-function Filter({ facets, onDismiss, isOpen = false }: Props) {
-  const { toggleFacets, state: searchState } = useSearch()
+function Filter({
+  facets,
+  onDismiss,
+  isOpen = false,
+  testId = 'store-filter',
+}: Props) {
+  const { width: screenWidth } = useWindowDimensions()
+  const { toggleFacet, toggleFacets, state: searchState } = useSearch()
 
   const [expandedIndices, setExpandedIndices] = useState<Set<number>>(
     new Set([])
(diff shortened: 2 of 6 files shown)


lucasfp13:
Replacement of this PR by #229 because we need to be based on `main-styled` branch.

src/components/search/Filter/Filter.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const [selectedFilters, setSelectedFilters] = useState<IStoreSelectedFacet[]>(
+    searchState.selectedFacets ?? []
+  )
+
+  const [isMobile, setIsMobile] = useState<boolean>(false)
+
+  useEffect(() => {
+    if (screenWidth) {
+      // notebook breakpoint = 1280px (See breakpoints on styles/global.scss)
+      setIsMobile(screenWidth < 1280)
```
hellofanny, line 61: Maybe we can create a variable for that on `src/constants.ts` file.

filipewl (approved):

src/components/search/Filter/Filter.tsx
hellofanny: I loved this idea <3

hellofanny (approved):
LGTM! 🚀 🚀

src/components/search/Filter/Filter.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const onCheck = ({ key, value }: IStoreSelectedFacet) => {
+    if (isMobile) {
+      onFilterChange({ key, value })
+
+      return
+    }
+
+    toggleFacet({ key, value })
+    onFilterChange({ key, value })
```
danzanzini, line 99:
```suggestion
    if (!isMobile) {
      toggleFacet({ key, value })
    }

    onFilterChange({ key, value })
```

src/components/search/Filter/Filter.tsx
filipewl: Oh, I see! We can make the switch later then.

filipewl:
I'm going to merge it so we can see if the preconnect will help the `main-styled` branch.

src/components/search/Filter/Filter.tsx
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  const Filters = () => {
```
danzanzini, line 102: When first seeing this I got confused about why `Filters` were inside the `Filter`. I thought `Filters` were a collection of `Filter`, but it's not 🤭.

Since there is no collection of filters thing, the confusion was a naming thing. I think the correct name for this const would be `Facets`, that is the options user have to narrow the search. More here: https://www.loop54.com/blog/what-is-a-search-facet-what-is-a-search-filter

danzanzini (commented):
Amazing work, @lucasfp13. The code looks great. 🎉 

I have two observations:
1) Regarding the behavior. When I'm at the `/office` collection page, the `office` facet is still available for being deselected, but when I do, nothing happens. I think it's strange to be able to deselect it when I'm at the `/office` page.

2) This PR adds a new feature that is not so simple. I know you are all in a rush but this deserved a better description, it would help a lot on the revision.

src/components/search/Filter/Filter.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const onCheck = ({ key, value }: IStoreSelectedFacet) => {
+    if (isMobile) {
+      onFilterChange({ key, value })
+
+      return
+    }
+
+    toggleFacet({ key, value })
+    onFilterChange({ key, value })
+  }
```
eduardoformiga, line 100: Not sure about that but since we depend on previous values should we refactor [to use a second form of setState()](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) that accepts a function rather than an object?

src/components/search/Filter/Filter.tsx
```diff
@@ -0,0 +1,185 @@
...
+      )
+
+      selectedFilters.splice(indexToRemove, 1)
+      setSelectedFilters([...selectedFilters])
+
+      return
+    }
+
+    setSelectedFilters([...selectedFilters, item])
+  }
```
eduardoformiga, line 89: Not sure about that but since we depend on previous values should we refactor [to use the second form of setState()](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) that accepts a function rather than an object on the 2 functions above?

gatsby-ssr.js
```diff
@@ -42,6 +42,12 @@ export const onRenderBody = ({ setHeadComponents }) => {
         enablePartytown
       />,
       <Partytown key="party" />,
+      <link
+        key="preconnect"
+        rel="preconnect"
+        crossOrigin
+        href="https://googletagmanager.com"
+      />,
```
danzanzini, line 50: I think this should have its own PR. It's not related to the Filter component.

src/components/search/Filter/Filter.tsx
```diff
@@ -0,0 +1,185 @@
...
+        X
+      </Button>
+      <Filters />
+    </UIModal>
+  ) : (
+    <>
+      <h4>Filters</h4>
+      <Filters />
+    </>
+  )
```
danzanzini, line 182: It seems strange to me that the component itself decides if it's going to be at a Modal or not. It looks like this should be its parent's responsibility. What do you think?

gatsby-ssr.js
hellofanny: We had https://github.com/vtex-sites/base.store/pull/233 but we merged this PR before.

PR #231, opened 2022-01-12
Fix: Adjust Data Layer has event

ArthurTriis1:
Signed-off-by: Arthur Andrade <arthurfelandrade@gmail.com>

Fix analytics events with intermittent errosr. Adding [cypress-wait-until](https://www.npmjs.com/package/cypress-wait-until) dependence and fixing dataLayerHasEvent

### DIFF
--- a/cypress/integration/analytics.test.js
+++ b/cypress/integration/analytics.test.js
@@ -15,11 +15,15 @@ beforeEach(() => {
 })
 
 const dataLayerHasEvent = (eventName) => {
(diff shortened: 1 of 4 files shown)


filipewl (approved):
Yay! Apparently, this does the trick to solve the following flaky test!

Flaky test

I'm in favor of merging this even if Lighthouse is not passing since the error is not caused by this PR.

eduardoformiga (approved):

PR #232, opened 2022-01-12
Feat: login (subdomain)

lariciamota:
Add the login feature to the base store.

## How it works? 
The login flow redirects to this subdomain `https://login.vtex-base1.tk/` which essentially corresponds to `storeframework.vtexcommercestable.com.br` 
In the end, it redirects back to the base store home page (using this domain: https://vtex-base1.tk/).

## How to test it?
Go to `/login` and follow the login flow.

## References
- We use [Freenom](https://my.freenom.com/) for DNS management
- A script was added in the old CMS in order to redirect `storeframework.vtexcommercestable.com.br`  back to the base store at the end of the flow.
   - To access the home layout in the old CMS: go to the [admin](https://storeframework.myvtex.com/admin/a/) and click `CMS -> HTML Templates -> Home`
   - The redirect is done using the meta tag `<meta http-equiv="refresh" content="0; URL='https://vtex-base1.tk/'"/>`
   - We also cleaned the script to contain only the redirect

## Next Steps
- Set cookie to all subdomains, we are in contact with the Identity team to have this implemented. ([This](https://vtex.slack.com/archives/C90SGTR6U/p1641510168025500?thread_ts=1640629695.217400&cid=C90SGTR6U) is one thread with discussion.)
- Redirect directly to myvtex instead of vtexcommercestable. ([This](https://vtex.slack.com/archives/C02R9LPL8EB/p1641823025018300?thread_ts=1640726107.047000&cid=C02R9LPL8EB) is a thread with discussion.)


src/pages/login.tsx
```diff
@@ -1,14 +1,16 @@
-import React from 'react'
+import React, { useEffect } from 'react'
```
igorbrasileiro, line 1: Does this login page really need to exist? A redirect doesn't work?

```ts
createRedirect({
  fromPath: '/login',
  toPath: loginUrl,
  statusCode: 301, // can try remove this status code and set `redirectInBrowser: true`
})
```

igorbrasileiro (approved):

src/pages/login.tsx
lariciamota: It could work but adds more complexity to the solution since it requires a lot more `createRedirects` as can be seen in [this commit](https://github.com/vtex-sites/base.store/pull/232/commits/c786e5323d6ea171e6e58ea291a01df6106d3e75) and is really tied to Gatsby. Besides that, something else is missing, maybe adding some header to the redirect requests could solve - you can see in [this preview](https://sfj-c786e53--base.preview.vtex.app/) that the login is throwing an error.
Although the redirect solution that changes the URL is slower ([this preview](https://sfj-25367af--base.preview.vtex.app/)), I'm leaning towards preferring it to the `createRedirect` one 🤔

emersonlaurentino: The advantage that using this approach with `useEffect` is the fact that we can customize the page with loading for example, but I don't know if this is enough for us to adopt. The second approach is more comfortable to me because we centralize the rules of redirects, but you had a good point when you said about the is tied to Gatsby, however, the gatsby will create this redirect on Nginx in our platform or in another thing in another platform. So, I'm leaning towards using the second approach for reason that we can separate the responsibility and the loading is not enough to keep the first approach.

Make sense for you @lariciamota ?

lariciamota: It does! I'll continue working on this approach then so it can be a viable solution for us

victorhmp (approved):

PR #241, opened 2022-01-14
Feat: Style filter

hellofanny:
## What's the purpose of this pull request?

- Styles Filter
- Styles Filter Button on mobile
- Adds `title-small` class on `typography.scss` Extract to https://github.com/vtex-sites/base.store/pull/245

|Desktop|Mobile|
|-|-|
Desktop|Mobile

To Do later:
- Styles Checkbox component
- Uses slide-in effect instead of Modal component

## How to test it?

https://preview-241--base.preview.vtex.app/office/

## References
[Jira issue FSSS-117 Style "Filter" component](https://vtex-dev.atlassian.net/browse/FSSS-117)

PR #244, opened 2022-01-17
Chore: remove unused css (spike) [FSSS-143]

eduardoformiga:
This PR aims to apply some experiments in order to deal with critical CSS, unused CSS, code-splitting.

Edit Note:
**The code from this PR was also added on f7ae9c8 on #241. so that we can solve the unused CSS problem there.
So this PR here becomes a study case that we can visit as necessary in the future :)**

### Context
Lighthouse (LH) started complaining about `unused css`:

> ✘ unused-css-rules failure for maxLength assertion
> Reduce unused CSS
> https://web.dev/unused-css-rules/

Nowadays:
- Using the build command (`yarn build`), the CSS from all pages and components are injected inside each page, minified, inside a `<style>` tag in the `<head>`. Thus, the browser avoids one more request to fetch a CSS file.

Screen Shot 2022-01-17 at 11 46 12

- Locally, using develop mode (`yarn develop`) everything is extracted into a CSS file (commons.css) and imported using `<link>` in the `<head>`. This function is the function of webpack's `mini-css-extract-plugin`.

Screen Shot 2022-01-17 at 13 27 24

### Problem
**All the CSS is being injected into pages globally, even unused ones.**

This is a known issue in Gatsby.
In the `v2` version, a technique called `code-splitting CSS` was used, I imagine that the CSS of each component/page was separate, and only what was used was imported into the pages. But this proved to be problematic [#11072](https://github.com/gatsbyjs/gatsby/issues/11072).

So in `v3`, which is the version we are using now, `code splitting` has become optional and needs to be done manually and all CSS is placed globally on each page as mentioned earlier. Thus, we need to at least scope the CSS so that we don't have either style breaking or unexpected behavior. This generates a lot of unused CSS on the pages. Also known issue [#28046](https://github.com/gatsbyjs/gatsby/issues/28046), [#21728](https://github.com/gatsbyjs/gatsby/issues/21728), [#18732 ](https://github.com/gatsbyjs/gatsby/issues/18732).

### Techniques and tools

The technique recommended by `web.dev`:

> Inline critical CSS and defer non-critical CSS

references: [#1](https://web.dev/extract-critical-css/), [#2](https://web.dev/unused-css-rules/#inline-critical-css-and-defer-non-critical-css), [#3](https://web.dev/defer-non-critical-css/)

Screen Shot 2022-01-21 at 18 43 31

There are some tools and plugins to do this. But we need to know if it matches well with the `gatsby`/`webpack` stack and test:
[critical](https://github.com/addyosmani/critical/blob/master/README.md), [html-critical-webpack-plugin](https://github.com/anthonygore/html-critical-webpack-plugin).

Another technique is to `remove unused CSS`, using something like [PurgeCSS](https://purgecss.com/), which also has [gatsby-plugin-purgecss](https://www.gatsbyjs.com/plugins/gatsby-plugin-purgecss/) for gatsby.

There is also another technique known as `tree-shaking` ([#4](https://linguinecode.com/post/reduce-css-file-size-webpack-tree-shaking), [#5](https://markmurray.co/blog/tree-shaking-css-modules/)).

Screen Shot 2022-01-21 at 18 45 04

[css-modules](https://github.com/css-modules/css-modules) doesn't solve our problem. I did some tests too and what it does is scope our classes well. It uses a hash for this. The final markup is a little worse to read but it solves the class collision problem. It might be a future approach, nowadays we scope using `BEM + data-attributes`. The `css-modules` is very class-focused, we would have to adapt the `data-attributes`, for example. It would solve the collision problem but not the unused CSS problem.

[This repository](https://github.com/MicheleBertoli/css-in-js) lists some `css-in-js` techniques. `styled-components` solves our problem, but I don't particularly like how the final hashed markup looks. We would have to adapt a lot of things in our CSS too.

I think it's worth doing some experiments and trying to adapt the rules of `webpack` from gatsby [like this](https://www.gatsbyjs.com/docs/how-to/custom-configuration/add-custom-webpack-config). Although understanding the [initial configuration](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/webpack.config.js) that gatsby does is a bit tricky.

### CSS splitting
commit: https://github.com/vtex-sites/base.store/pull/244/commits/5dc155b12a10dd747be6d205ec35afdee8e0e7de
CSS coverage after this Webpack setup:
|Before|After|
|-|-|
|Screen Shot 2022-01-17 at 15 58 56|Screen Shot 2022-01-17 at 16 53 24|

✅  Code splitting CSS with only the global CSS and the page CSS. Does not include the other page's CSS.

### CSS splitting + Critical CSS
commit: https://github.com/vtex-sites/base.store/pull/244/commits/9d00e9715102f2c86c69039b29d3adb5ca43fce8
Testing [critical](https://github.com/addyosmani/critical/blob/master/README.md) CSS only on homepage.
inline `critical above the fold CSS` and defer `uncritical below the fold css`

**Result**: it's not possible to test in the preview environment since we need the puppeteer installed.

❌ **Preview Build** failed:
```
error UNHANDLED REJECTION Failed to launch the browser process! spawn /workspace/build/chore/remove-unused-css/vtex-sites/base.store/node_modules/puppeteer/.local-chromium/linux-722234/chrome-linux/chrome ENOENT
```

**Locally**:

Screen Shot 2022-01-19 at 21 16 19

❌  Drawback:
- Lots of unused CSS but only in external CSS files (inside `<link rel="stylesheet">`). 
- Inlined CSS is not cached.
- Lots of manual configs. We need to apply this for all static files for instance and we don't know beforehand all the filenames.
- We need to install the puppeteer in CI.

✅  The critical CSS was applied inline and has a good CSS used rate, but some duplicated CSS needs to be removed yet (from the page's CSS).

### CSS splitting + PurgeCSS

commit: https://github.com/vtex-sites/base.store/pull/244/commits/8b08fd343e892e7668b3a7ed396262dc370ad4b0

Screen Shot 2022-01-20 at 13 40 29

❌  PurgeCSS does not support attribute (like `data-attributes`) or complex selectors: Issue [#110](https://github.com/FullHuman/purgecss/issues/110).
✅ When we just purge the `global.scss` we gain almost nothing +- (0,5kb).

### CSS splitting + Custom Loaders
❌ Attempts to optimize our CSS with custom loaders like [dropCSS](https://github.com/leeoniya/dropcss), [clean-css](https://github.com/clean-css/clean-css), [CSSO](https://github.com/sandark7/csso-loader), but the final result doesn't have much difference in the final result.

### CSS splitting + Tree shaking.
commit: https://github.com/vtex-sites/base.store/pull/244/commits/8f1b57d7dda0232e5a29191458ad936fc30b81f1

❌  Testing with `sideEffects` as true, false, and using either `loaders` or inside the `package.json` broke the style or doesn't have much difference in the final result.

## Conclusion

Let's just keep the code splitting CSS with scoped CSS so far. This technique brings more gains without cumbersome other configs/techniques that bring less for our context.


eduardoformiga:
The code from this PR was also added on https://github.com/vtex-sites/base.store/pull/241/commits/f7ae9c80c580b2074c2ebed6cbc2c91ca41a0d5b on https://github.com/vtex-sites/base.store/pull/241. so that we can solve the unused CSS problem there. 
So this PR here becomes a study case that we can visit as necessary in the future :)

PR #249, opened 2022-01-18
fix: change cypress version

ArthurTriis1:
Signed-off-by: Arthur Andrade <arthurfelandrade@gmail.com>

Currently cypress runs in two different versions on local development and on the CI. This PR fix the base.store version to '6.6.0'

### DIFF
--- a/src/sdk/analytics/platform/vtex.ts
+++ b/src/sdk/analytics/platform/vtex.ts
@@ -1,5 +1,5 @@
 import type { AnalyticsEvent } from '@faststore/sdk'
 
 export default function sendEvent(event: AnalyticsEvent) {
-  window.sendrc(event.name, event.params)
+  window.sendrc?.(event.name, event.params)
 }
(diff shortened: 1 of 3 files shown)


src/sdk/analytics/platform/vtex.ts
```diff
@@ -1,5 +1,5 @@
 import type { AnalyticsEvent } from '@faststore/sdk'
 
 export default function sendEvent(event: AnalyticsEvent) {
-  window.sendrc(event.name, event.params)
+  window.sendrc?.(event.name, event.params)
```
eduardoformiga, line 4: Is it related to the cypress?

eduardoformiga (approved):

src/sdk/analytics/platform/vtex.ts
ArthurTriis1: Isn't, but have broken when i changed the versions

PR #247, opened 2022-01-18
Chore: fix CLS Lighthouse

eduardoformiga:
## What's the purpose of this pull request?
This PR attempts to fix the Cumulative Layout Shift CLS from the Lighthouse CI.

PR #240, opened 2022-01-14
[FSSS-121] Feat: Style the PLP

filipewl:
## What's the purpose of this pull request?

Style the PLP according to [this Figma file](https://www.figma.com/file/wS0Kek1aSzIRl20jdtnSsL/Base-%E2%80%93-Closed-Beta-Version?node-id=884%3A88434).

Expectation|Reality
-|-
|

Expectation|Reality
-|-
|
 
## How to test it?
Visit the [deploy preview](https://preview-240--base.preview.vtex.app/office).

## References
- [Jira FSSS-121 Style PLP Page](https://vtex-dev.atlassian.net/browse/FSSS-121)

PR #250, opened 2022-01-18
Fix: Attempt to fix PLP Lighthouse CLS issue

filipewl:

Another attempt to fix PLP's Cumulate Layout Shift plaguing `main-style` and its other branches.

## How does it work? 

On the PLP page load, the footer starts occupying a big part of the viewport and it has some internal layout shifts in the incentives list area, as well as the transformation of the footer links into the accordion.

Incentives|Accordion
-|-
incentives|accordion

This temporary fix increases the minimum height of the main element so it pushes the footer the minimum amount to be below the fold. That is possible because the height of the alert and navbar don't change frequently. It's believed that when https://github.com/vtex-sites/base.store/pull/240 lands, we won't need this fix and can remove it. Or when we implement page skeletons in the future.

## How to test it?

The Lighthouse CI check passes.

Before|After
-|-
Before|After

## References

- Another attempt to fix this is at https://github.com/vtex-sites/base.store/pull/247 but the CLS still failed in one of the commits.

### DIFF
--- a/src/Layout.tsx
+++ b/src/Layout.tsx
@@ -47,7 +47,7 @@ function Layout({ children }: PropsWithChildren<unknown>) {
 
       <Navbar />
 
-      <main>{children}</main>
+      <main className="temp-cls-fix">{children}</main>
 
       <Footer />
       {displayMinicart && (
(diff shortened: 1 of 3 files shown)


hellofanny (commented):
👏 👏 👏

PR #252, opened 2022-01-18
feat: Create Accordion component

lucasfp13:

This PR intends to add the `Accordion` component and its child component `AccordionItem`. Both were created using FastStore's Accordion component, but in a more straightforward way with the addition of some props to achieve the same result.

Internally, the `AccordionItem` uses the `AccordionButton` and `AccordionPanel` components as currently we only have one example of using the accordion.

## How it works? 

Here is an example of how it can be used:

Screen Shot 2022-01-18 at 14 24 51


src/components/ui/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { forwardRef } from 'react'
+import type { HTMLAttributes } from 'react'
+import {
+  Icon as UIIcon,
+  AccordionItem as UIAccordionItem,
+  AccordionPanel as UIAccordionPanel,
+  AccordionButton as UIAccordionButton,
+} from '@faststore/ui'
+
+interface Props extends HTMLAttributes<HTMLDivElement> {
```
filipewl, line 10: I wonder if it's a good idea to extend from `AccordionItemProps` (imported from `@faststore/ui`)?

Same comment for `Accordion`'s and `AccordionProps` https://github.com/vtex-sites/base.store/blob/6b872ab34aad9927eff221930f00f1791fe18cd7/src/components/ui/Accordion/Accordion.tsx#L5

filipewl (approved):
👏

src/components/ui/Accordion/AccordionItem.tsx
lucasfp13: Good idea, thanks!!

danzanzini: It seems like a good idea to me. Very straightforward.

danzanzini (approved):
:shipit:

src/components/search/Filter/Filter.tsx
```diff
@@ -102,47 +95,45 @@ function Filter({
...
+                      </li>
+                    )
+                  })}
+                </UIList>
+              </AccordionItem>
             ))}
-        </UIAccordion>
+        </Accordion>
         {isMobile && (
           <div>
```
hellofanny, line 138:
```suggestion
          <footer>
```

eduardoformiga (approved):
LGTM! 👏

PR #255, opened 2022-01-19
fix: error accessing price and identifier when build product and cate…

birnayt:
<em>Fix the error of trying to access the price and identifier when the API does not return these values.</em>

## How it works? 
<em>Just put a optional chaining in places where the value doesn't always exist </em>

## How to test it?
<em>Try using a marketplace like "homologappoffpremium" without that fix and run the build two or three times.</em> 

erro-build

### DIFF
--- a/src/components/product/ProductCard/ProductCard.tsx
+++ b/src/components/product/ProductCard/ProductCard.tsx
@@ -41,7 +41,9 @@ function ProductCard({ product, index, className }: Props) {
     return lowestPriceOffer
   }, [spotPrice, offers])
 
-  const { listPrice, seller } = offers[selectedOffer]
+  const selectedOfferPrice = offers[selectedOffer]
+  const listPrice = selectedOfferPrice?.listPrice
+  const seller = selectedOfferPrice?.seller
 
   const linkProps = useProductLink({ product, selectedOffer, index })
   const buyProps = useBuyButton({
--- a/src/components/sections/ProductDetails/ProductDetails.tsx
+++ b/src/components/sections/ProductDetails/ProductDetails.tsx
@@ -34,11 +34,13 @@ function ProductDetails({ product: staleProduct }: Props) {
       brand,
       isVariantOf,
       image: productImages,
-      offers: {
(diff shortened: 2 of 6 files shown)


icazevedo (approved):
LGTM. This doesn't fix the root of the problem but while we don't find a more permanent fix to the issue, we shouldn't have problems building the store.

igorbrasileiro (approved):

victorhmp (approved):

src/components/sections/ProductDetails/ProductDetails.tsx
```diff
@@ -34,13 +34,13 @@ function ProductDetails({ product: staleProduct }: Props) {
...
+      offers: { offers },
     },
   } = data
 
-  const price = offers.offers[0]?.price
-  const listPrice = offers.offers[0]?.listPrice
-  const seller = offers.offers[0]?.seller
+  const price = offers[0]?.price
+  const listPrice = offers[0]?.listPrice
+  const seller = offers[0]?.seller
```
igorbrasileiro, line 43: Can you explain the motivation behind those changes? IMO, it seems to be the same, semantically and for readability.

igorbrasileiro (changes requested):

src/components/sections/ProductDetails/ProductDetails.tsx
birnayt: The motivation was just to improve reading.

src/components/product/ProductCard/ProductCard.tsx
```diff
@@ -41,7 +41,9 @@ function ProductCard({ product, index, className }: Props) {
     return lowestPriceOffer
   }, [spotPrice, offers])
 
-  const { listPrice, seller } = offers[selectedOffer]
+  const selectedOfferPrice = offers[selectedOffer]
+  const listPrice = selectedOfferPrice?.listPrice
+  const seller = selectedOfferPrice?.seller
```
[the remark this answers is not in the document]
birnayt: Hmmm thanks, didn't know about this one. I will fix.

PR #268, opened 2022-01-24
Remove dup adding/removing cart items in the PLP tests block

filipewl:

These actions are already being tested in the [PDP describe block](https://github.com/vtex-sites/base.store/blob/4c830bb2c0fac98bf41214f3a3cd030370e28cbe/cypress/integration/cart.test.js#L28-L82). The PLP ones just repeated the same process by visiting a PDP. It made sense before when you could add a product to the cart on the PLP, but now this action can only be done on the PDP.


eduardoformiga (approved):

israelswf (approved):
🆗 ✨

PR #269, opened 2022-01-24
chore: Filter component's fixes

lucasfp13:

This PR intends to fix `Filter`s wrong behaviors, mainly for the mobile version. For instance, the component wasn't clearing all facets when requested or applying the right facets.

- [x] Keep page slug option always checked and disabled;
- [x] Handle multiple check and unchecking facets;
- [x] Fix clearing all facets;

### DIFF
--- a/src/components/search/Filter/Filter.tsx
+++ b/src/components/search/Filter/Filter.tsx
@@ -206,7 +206,11 @@ function Filter({
     )
   }
 
-  return isMobile ? (
+  if (!isMobile) {
+    return <Facets />
+  }
+
+  return (
     <SlideOver
       isOpen={isOpen}
       onDismiss={onDismiss}
@@ -265,8 +269,6 @@ function Filter({
         </Button>
       </footer>
     </SlideOver>
-  ) : (
-    <Facets />
   )
 }
 
--- a/src/components/search/Filter/Filter.tsx
+++ b/src/components/search/Filter/Filter.tsx
@@ -87,9 +87,12 @@ function Filter({
     [indicesExpanded]
   )
 
-  // Ensures selected filters are up to date at opening
+  // Ensures all required states are up to date at opening
   useEffect(() => {
-    !isOpen && setSelectedFacets(searchState.selectedFacets)
+    if (!isOpen) {
+      setFacetsToRemove([])
+      setSelectedFacets(searchState.selectedFacets)
+    }
   }, [isOpen, searchState.selectedFacets])
 
   // Opens accordion items with active facets
@@ -118,9 +121,12 @@ function Filter({
         (f) => f.value === item.value
       )
 
+      if (selectedFacets.some((facet) => facet.value === item.value)) {
+        setFacetsToRemove([...facetsToRemove, item])
+      }
+
       selectedFacets.splice(indexToRemove, 1)
       setSelectedFacets([...selectedFacets])
-      setFacetsToRemove([...facetsToRemove, item])
 
       return
     }
@@ -240,26 +246,22 @@ function Filter({
           variant="primary"
           data-testid="filter-modal-button-apply"
           onClick={() => {
-            // Remove facets that user unchecked
-            facetsToRemove.length > 0 && toggleFacets(facetsToRemove)
-
             // Only toggle new facets and keep the current ones applied
             const facetsToAdd = selectedFacets
               .map(
                 (facet) => !searchState.selectedFacets.includes(facet) && facet
               )
+              .concat(facetsToRemove)
               .filter((m) => typeof m !== 'boolean') as IStoreSelectedFacet[]
 
             toggleFacets(facetsToAdd)
 
             setActiveFacets([])
-            setFacetsToRemove([])
-            setSelectedFacets([])
             setIndicesExpanded(new Set([]))
             onDismiss?.()
           }}
         >
-          View Results
+          Apply
         </Button>
       </footer>
     </SlideOver>
--- a/src/components/search/Filter/Filter.tsx
+++ b/src/components/search/Filter/Filter.tsx
@@ -100,12 +100,12 @@ function Filter({
     }
 
     // Ensures there isn't empty facets
-    const filteredActiveFacets = activeFacets.filter(
+    const selectedActiveFacets = activeFacets.filter(
       (item) => item.facets.length > 0
     )
 
     // Checks if accordion item is already opened
-    filteredActiveFacets.map(
+    selectedActiveFacets.map(
       ({ accordionIndex }) =>
         !indicesExpanded.has(accordionIndex) &&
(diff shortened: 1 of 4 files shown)


src/components/search/Filter/Filter.tsx
```diff
@@ -60,16 +92,30 @@ function Filter({
...
+    if (activeFacets.length !== filteredFacets.length) {
       return
     }
 
-    setIndicesExpanded(new Set(indicesExpanded.add(index)))
-  }
+    // Ensures there isn't empty facets
+    const filteredActiveFacets = activeFacets.filter(
+      (item) => item.facets.length > 0
+    )
```
danzanzini, line 110: I'm confused regarding the concepts of filtered and active facets. Can you clarify?

src/components/search/Filter/Filter.tsx
```diff
@@ -60,16 +92,30 @@ function Filter({
     }
   }, [screenWidth])
 
-  const onAccordionChange = (index: number) => {
-    if (indicesExpanded.has(index)) {
-      indicesExpanded.delete(index)
-      setIndicesExpanded(new Set(indicesExpanded))
+  // Ensures selected filters are up to date at opening
+  useEffect(() => {
+    !isOpen && setSelectedFilters(searchState.selectedFacets)
```
danzanzini, line 97: Can `setSelectedFilters` be renamed to `setSelectedFactes`?

src/components/search/Filter/Filter.tsx
```diff
@@ -50,8 +57,33 @@ function Filter({
     searchState.selectedFacets ?? []
   )
 
-  const [isMobile, setIsMobile] = useState<boolean>(false)
+  const [filtersToRemove, setFiltersToRemove] = useState<IStoreSelectedFacet[]>(
+    []
+  )
+
```
danzanzini, line 63: `facetsToRemove`?

danzanzini (commented):
Good job, @lucasfp13. It works great, now.

I made an initial review of the code and left some suggestions. There are places where `filters` is still used, maybe it should be changed to `facets`

src/components/search/Filter/Filter.tsx
```diff
@@ -50,8 +57,33 @@ function Filter({
...
   let onDismissTransition: () => unknown
+  const [isMobile, setIsMobile] = useState<boolean>(false)
+  const [activeFacets, setActiveFacets] = useState<ActiveFacets[]>([])
```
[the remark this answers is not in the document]
lucasfp13: yes, I totally agree with you, but the Prettier doesn't hahaahha when the line gets too big and is broken, Prettier waits for a space before and after

src/components/search/Filter/Filter.tsx
lucasfp13: This filtering intends to prevent the array to have empty values (happens when there is no checked facets into the accordion item), so is basically to avoid unnecessary checks in empty arrays

src/components/search/Filter/Filter.tsx
```diff
@@ -170,8 +238,9 @@ function Filter({
         <Button
           variant="secondary"
           onClick={() => {
-            toggleFacets(selectedFilters)
-            setIndicesExpanded(new Set([]))
+            const filters = selectedFilters
+
+            setFiltersToRemove(filters)
```
ArthurTriis1, line 243: Why not?
```suggestion
            setFiltersToRemove(selectedFilters)
```

src/components/search/Filter/Filter.tsx
```diff
@@ -102,43 +149,61 @@ function Filter({
...
+                const filteredValues = values.filter(({ selected }) => selected)
+
+                // Ensures only one array item for each accordion's item
+                if (activeFacets.length < filteredFacets.length) {
+                  activeFacets.push({
+                    accordionIndex: index,
+                    facets:
+                      filteredValues.length > 0
+                        ? filteredValues.map(({ value }) => value)
+                        : [],
```
danzanzini, line 169: I think this ternary is not necessary. If you map an empty array you will get an empty array.

src/components/search/Filter/Filter.tsx
```diff
@@ -102,43 +149,61 @@ function Filter({
...
-                          checked={selectedFilters.some(
+          {filteredFacets.map(({ label, values, key }, index) => (
+            <AccordionItem
+              key={`${label}-${index}`}
+              testId="filter-accordion"
+              isExpanded={indicesExpanded.has(index)}
+              buttonLabel={label}
+              ref={(_) => {
+                // Filter current selected facets from API
+                const filteredValues = values.filter(({ selected }) => selected)
```
danzanzini, line 160: `const selectedValues`

src/components/search/Filter/Filter.tsx
danzanzini: Got it. I think we can improve these names to clarify the code.

Regarding to facets, we have:
- `filteredFacets`
- `filteredActiveFacets`

These ones can be better named by telling what the filter does. i.e., if you're filtering the selected `activeFacets` the name should be `selectedActiveFacets`.

filipewl:
Nice improvements!!!

However, there seems to exist something weird when checking & unchecking multiple filters. Sometimes they just won't work.

Recorded the following steps:
CleanShot 2022-01-26 at 12 28 55

src/components/search/Filter/Filter.tsx
```diff
@@ -181,8 +248,23 @@ function Filter({
...
+
+            // Only toggle new facets and keep the current ones applied
+            const facetsToToggle = selectedFacets
+              .map(
+                (filter) =>
+                  !searchState.selectedFacets.includes(filter) && filter
+              )
+              .filter((m) => typeof m !== 'boolean') as IStoreSelectedFacet[]
+
+            toggleFacets(facetsToToggle)
```
heitorado, line 262: What do you think about some renaming to improve readability/semantics? Here's a suggestion:
```suggestion
            const facetsToAdd = selectedFacets
              .map(
                (facet) =>
                  !searchState.selectedFacets.includes(facet) && facet
              )
              .filter((m) => typeof m !== 'boolean') as IStoreSelectedFacet[]

            toggleFacets(facetsToAdd)
```

saranicoly:
When running on mobile, if I apply some filter the page doesn't scroll anymore:

https://user-images.githubusercontent.com/22377378/151402353-e4d754de-ec47-4555-b0f6-b1055dbc5728.mov

saranicoly (commented):
When running on mobile, if I apply some filter the page doesn't scroll anymore:

https://user-images.githubusercontent.com/22377378/151403471-6e64c8d2-4b03-4d53-baeb-3a046c8fbcd5.mov

saranicoly (approved):
Great job!

PR #280, opened 2022-01-26
fix: allows for GTM debugging

icazevedo:
This PR makes GTM debugging work properly.

## How it works? 
Since the debugging script GTM uses breaks inside partytown, we're using the `gtm_debug` query string as a flag for enabling/disabling partytown.

## How to test it?
Include the gtm_debug query string on any page of the website and check if the GTM script loads outside partytown (no text/partytown type on the script tag).

## References
We've tried to make it work using partytown directly, but with no success. The partytown team seems to be fixing the issues.


igorbrasileiro:
@icazevedo can you update this branch?!

src/sdk/analytics/index.tsx
```diff
@@ -1,15 +1,22 @@
...
 }
 
-export const AnalyticsHandler = ({ children }: PropsWithChildren<unknown>) => {
+const GTM_DEBUG_QUERY_STRING = 'gtm_debug'
+
+export const AnalyticsHandler = () => {
   useAnalyticsEvent((event: AnalyticsEvent) => {
+    // Cleans the ecommerce object before pushing a new one
+    // This prevents the new data from getting perged with the previous one
+    // which could lead do inacurate and old data being sent with events
```
igorbrasileiro, line 18: It would be good if you left the reference that says it.

src/sdk/analytics/index.tsx
```diff
@@ -19,7 +26,17 @@ export const AnalyticsHandler = ({ children }: PropsWithChildren<unknown>) => {
     )
   })
 
-  return children
+  if (typeof window === 'undefined') {
+    return null
```
igorbrasileiro, line 30:
```suggestion
    return children
```

When the component returns null all the subtree isn't rendered. So doing this, the SSG won't work as expected

igorbrasileiro (changes requested):

icazevedo:
We'll need to stay on version 2.4 until https://github.com/BuilderIO/partytown/pull/68 gets merged

src/sdk/analytics/index.tsx
```diff
@@ -1,15 +1,43 @@
...
+  `${
+    opts.partytownScript ? '!' : ''
+  }window.location.search.includes('${GTM_DEBUG_QUERY_STRING}=')&& 
+  (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+ new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+ 'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+ })(window,document,'script',${JSON.stringify(
+   opts.dataLayerName ?? 'dataLayer'
+ )},${JSON.stringify(opts.containerId)});`
```
igorbrasileiro, line 31: To be honest, this code lgtm, but i think it should be inside another file, something like a `src/sdk/analytics/GoogleTagManager`. What do you think about

gatsby-ssr.js
```diff
@@ -43,11 +40,37 @@ export const onRenderBody = ({ setHeadComponents }) => {
...
         key="gtm"
-        containerId={storeConfig.analytics.gtmContainerId}
-        enablePartytown
+        type="text/javascript"
+        dangerouslySetInnerHTML={{
+          __html: googleTagManager({
+            containerId: storeConfig.analytics.gtmContainerId,
+            partytownScript: false,
+          }),
+        }}
```
igorbrasileiro, line 73: Have you ever thought about adding only 1 script?

I thought to put the snippet inside a template tag and in the browser, put the template content into a script tag, but idk if it will harm the performance.

src/sdk/analytics/index.tsx
icazevedo: You're right. I'll move it.

gatsby-ssr.js
icazevedo: I've never thought of it, but I think it may harm performance and it makes the code more complex. In the end, I think the amount of code that'd be downloaded would be the same, so I don't feel like implementing it.

PR #293, opened 2022-01-31
[FSSS-170] Chore: tweaks search page

eduardoformiga:
Makes the Search page looks like the PLP so far.

**Question**: The search page is without any `<h1>` tag, should we add it someplace?
We decided to keep the `<h1>`, but use the `SROnly` (Screen Reader Only) component.

## How to test it?
https://preview-293--base.preview.vtex.app/s/?q=shirt&sort=score_desc&page=0


eduardoformiga:
Depends on https://base.vtex.app/s/?q=shirt to get the new theme/global styles.

filipewl (approved):
LGTM! 

Unfortunately, I have no clue how we can decrease the TTI. :(

src/components/ui/SROnly/SROnly.tsx
```diff
@@ -1,13 +1,17 @@
...
 import './sr-only.scss'
 
 interface Props {
   text: string
+  as?: ElementType
 }
 
-function SROnly({ text }: Props) {
-  return <span data-store-sr-only>{text}</span>
+function SROnly({ text, as }: Props) {
```
hellofanny, line 11: In the future we can rename it to a more generic meaning to cover this case (when we need to hide information but not necessary related to screen reader only)  👍

hellofanny (approved):
Nice job! 🤩

PR #295, opened 2022-02-01
chore: update login button

lariciamota:
Update the login button to redirect to login if the user isn't authenticated or to redirect to my account if it is.

## How it works? 
It gets the user info from session and updates the button accordingly. 
The faststore update that enables this get from session is in [this PR](https://github.com/vtex/faststore/pull/1139). 

## How to test it?
The login flow consists in redirecting to `vtexcommercestable` first (for more information about those integrations, see our [docs](https://faststore.dev/how-to-guides/platform-integration/vtex/hosting-a-faststore-vtex-website)) in a subdomain, when the login is completed the auth cookie is set on the root domain also - in the case of `base.store` the root domain is `vtexfaststore.com`. It means that when you test the login flow in a preview you'll get redirected back to https://vtexfaststore.com/ after the login and if you access the preview again you won't be authenticated because the cookie wasn't set there.
As a workaround, you can get the `VtexIdclientAutCookie_ACCOUNT` (eg `VtexIdclientAutCookie_storeframework`) and add it on the preview, for that you can use the [EditThisCookie](https://www.editthiscookie.com/) extension.

https://user-images.githubusercontent.com/19983991/157094151-ac065096-f0c3-4961-9b32-5e420da6e54b.mov


src/sdk/person/usePersonQuery.ts
```diff
@@ -0,0 +1,44 @@
...
+      id
+      email
+      givenName
+      familyName
+      isAuthenticated
+    }
+  }
+`
+
+export const usePersonQuery = (options?: QueryOptions) => {
```
igorbrasileiro, line 24:
```suggestion
const usePersonQuery = (options?: QueryOptions) => {
```

src/sdk/person/usePersonQuery.ts
```diff
@@ -0,0 +1,44 @@
...
+  const person = data?.person
+
+  useEffect(() => {
+    if (person !== null && person !== undefined && person !== user) {
+      setSession({
+        ...session,
+        user: person,
+      })
+    }
+  })
```
igorbrasileiro, line 41:
```suggestion
```suggestion
  useEffect(() => {
    if (person !== null && person !== undefined && person !== user) {
      setSession({
        ...session,
        user: person,
      })
    }
  }, [person, setSession, session])
```
```

File: src/sdk/person/usePersonQuery.ts (hunk 1 above)
Comment (igorbrasileiro) [reply src/sdk/person/usePersonQuery.ts:24]: Export default usePersonQuery

File: src/sdk/person/usePersonQuery.ts [hunk 3]
```diff
@@ -0,0 +1,46 @@
...
+    query,
+    {},
+    { ...options, fetchOptions: { ...options?.fetchOptions, method: 'POST' } }
+  )
+
+  const { setSession, user, ...session } = useSession()
+  const person = data?.person
+
+  useEffect(() => {
+    if (person !== null && person !== undefined && person !== user) {
```
igorbrasileiro, line 35:
```suggestion
    if (Boolean(person) && person !== user) {
```

Why person and user can't be equal? Also, remember the strict inequality (!==) will return true for a different object, even they have the same structure and same values

```js
{} !== {} // true
```

src/sdk/person/usePersonQuery.ts
```diff
@@ -0,0 +1,46 @@
...
+  query PersonQuery {
+    person {
+      id
+      email
+      givenName
+      familyName
+      isAuthenticated
+    }
+  }
+`
```
igorbrasileiro, line 22: Put it at the end of the file, before the export default, I think this is the pattern

src/sdk/person/usePersonQuery.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const usePersonQuery = (options?: QueryOptions) => {
+  const { data } = useQuery<PersonQueryQuery, PersonQueryQueryVariables>(
+    query,
+    {},
+    { ...options, fetchOptions: { ...options?.fetchOptions, method: 'POST' } }
```
igorbrasileiro, line 28: Does this request need to be a POST? I'm asking because the SWR (stale-while-revalidate) has no effect with POST.

src/sdk/person/usePersonQuery.ts
igorbrasileiro: Probably you wanna do a shallow equal, right?
Take a look at this shallowEqual from react.
https://github.com/facebook/react/blob/55d75005bc26aa41cddc090273f82aa106729fb8/packages/shared/shallowEqual.js

igorbrasileiro (changes requested):
Great job, just small tips

src/sdk/person/usePersonQuery.ts
victorhmp: I don't think this is standard in this repo @igorbrasileiro. The other SDK hooks that include GraphQL queries are also defining them first

src/sdk/person/usePersonQuery.ts
victorhmp: This has to be a POST just because it's sending the request to a GraphQL endpoint, isn't it?

victorhmp (approved):

src/sdk/person/usePersonQuery.ts
lariciamota: Actually, it has to be a POST to avoid caching

src/sdk/person/usePersonQuery.ts
lariciamota: Using `Boolean(person)` it was complaining as it wasn't checking if it was `undefined`, so I changed to check if the value is truth using `!!`, okay?
Screen Shot 2022-03-14 at 11 19 13

src/sdk/person/usePersonQuery.ts
igorbrasileiro: Ok, great

tlgimenes (approved):
LGTM 🎉

src/sdk/person/usePersonQuery.ts
lariciamota: > Why person and user can't be equal?
If they are equal then we wouldn't need to update `user`.
> Probably you wanna do a shallow equal, right?
Yeah, you are right! But for some reason it's doing what it should: if I remove the `person !== user` it enters an infinite user-update loop, but if I put this it only updates once - as it should. I don't know exactly why, but since it already interrupts the infinite loop I believe we don't need to implement the shallow equal. What do you think?

PR #299, opened 2022-02-02
[FSSS-164] Extract UISelect from Sort to its own component

heitorado:
This PR extracts the UISelect component that was being used by the Sort component internally to its own component.

## How it works? 
This essentially:
- Removes the `sort.scss` stylesheet completely
- Move all styles to the `select.scss` stylesheet
- Adds a new `Select` component that wraps the `UISelect` from faststore
  - It uses the `id` attribute to link the `label` component with the `UISelect`
  - If it receives a `className` attribute, it will forward that attribute to the `className` prop in the outer div. Optional.
  - If it receives a `labelText` attribute, it will render a label together with the select. Optional.
  - It forwards the `aria-label` attribute to the `UISelect` `aria-label`. This attribute is required.
  - It receives an `options` attribute which is a `Record<string, string>` so it can build the available options that will be displayed when the user clicks the Select component. This attribute is required.

## How to test it?
`yarn test`
Verify that all the tests still pass, so the original behaviour is kept.

Also open the local instance and verify that no styles are broken, meaning the change did not affect the styles as well.

## References
Thanks to @renatamottam

### DIFF
--- a/src/components/search/Sort/Sort.tsx
+++ b/src/components/search/Sort/Sort.tsx
@@ -23,7 +23,7 @@ function Sort() {
 
   return (
     <Select
-      className="sort / title-small"
+      classes="sort / title-small"
       labelText="Sort by"
       options={OptionsMap}
       onChange={(e) => setSort(keys[e.target.selectedIndex])}
--- a/src/components/search/Sort/Sort.tsx
+++ b/src/components/search/Sort/Sort.tsx
@@ -28,7 +28,6 @@ function Sort() {
       options={OptionsMap}
       onChange={(e) => setSort(keys[e.target.selectedIndex])}
       value={sort}
-      aria-label="Product Sort"
       testId="search-sort"
     />
   )
--- a/src/components/search/Sort/Sort.tsx
+++ b/src/components/search/Sort/Sort.tsx
@@ -23,6 +23,7 @@ function Sort() {
 
   return (
     <Select
+      id="sort-select"
       classes="sort / title-small"
       labelText="Sort by"
       options={OptionsMap}
--- a/src/components/search/Sort/Sort.tsx
+++ b/src/components/search/Sort/Sort.tsx
@@ -1,9 +1,6 @@
 import { useSearch } from '@faststore/sdk'
 import React from 'react'
-import { Select as UISelect } from '@faststore/ui'
-import { CaretDown as CaretDownIcon } from 'phosphor-react'
-
-import './sort.scss'
+import Select from 'src/components/ui/Select'
 
 const OptionsMap = {
   price_desc: 'Price, descending',
@@ -25,23 +22,15 @@ function Sort() {
   } = useSearch()
 
   return (
-    <div className="sort / title-small">
-      <label htmlFor="select-sort">Sort by</label>
-      <UISelect
-        data-testid="search-sort"
-        onChange={(e) => setSort(keys[e.target.selectedIndex])}
-        value={sort}
-        aria-label="Product Sort"
-        id="select-sort"
-      >
-        {keys.map((key) => (
-          <option key={key} value={key}>
-            {OptionsMap[key]}
-          </option>
-        ))}
-      </UISelect>
-      <CaretDownIcon size={18} weight="bold" />
-    </div>
+    <Select
+      className="sort / title-small"
+      labelText="Sort by"
+      options={OptionsMap}
+      onChange={(e) => setSort(keys[e.target.selectedIndex])}
+      value={sort}
+      aria-label="Product Sort"
+      testId="search-sort"
+    />
   )
 }
 
--- a/src/components/search/Sort/Sort.tsx
+++ b/src/components/search/Sort/Sort.tsx
@@ -24,7 +24,7 @@ function Sort() {
   return (
     <Select
       id="sort-select"
-      classes="sort / title-small"
+      className="sort / title-small"
       labelText="Sort by"
       options={OptionsMap}
       onChange={(e) => setSort(keys[e.target.selectedIndex])}
--- a/src/components/ui/Select/Select.tsx
+++ b/src/components/ui/Select/Select.tsx
@@ -12,10 +12,6 @@ interface UISelectProps extends SelectProps {
    */
   id: string
   /*
-   * Defines the CSS class string that will be forwarded to the wrapping div "className" prop.
-   */
-  classes?: string
-  /*
    * Defines the options available in the select. The SelectOptions object
    * keys are the property names, while the values correspond to the text that
    * will be displayed in the UI
@@ -30,16 +26,16 @@ interface UISelectProps extends SelectProps {
 
 export default function Select({
   id,
+  className,
   options,
   onChange,
   labelText,
   value,
   'aria-label': ariaLabel,
-  classes,
   testId,
 }: UISelectProps) {
   return (
-    <div data-select className={classes}>
+    <div data-select className={className}>
       {labelText && <label htmlFor={id}>{labelText}</label>}
       <UISelect
         data-testid={testId}
--- a/src/components/ui/Select/Select.tsx
+++ b/src/components/ui/Select/Select.tsx
@@ -16,7 +16,7 @@ interface UISelectProps extends SelectProps {
   classes?: string
   /*
    * Defines the options available in the select. The SelectOptions object
-   * keys are the property names, while the values correspond  to the text that
+   * keys are the property names, while the values correspond to the text that
    * will be displayed in the UI
    */
   options: SelectOptions
--- a/src/components/ui/Select/Select.tsx
+++ b/src/components/ui/Select/Select.tsx
@@ -11,6 +11,10 @@ type SelectOptions = {
 
 interface UISelectProps extends SelectProps {
   /*
+   * Defines the CSS class string that will be forwarded to the wrapping div "className" prop.
+   */
+  classes?: string
+  /*
    * Defines the options available in the select. The SelectOptions object
    * keys are the property names, while the values correspond  to the text that
    * will be displayed in the UI
@@ -29,11 +33,11 @@ export default function Select({
   labelText,
   value,
   'aria-label': ariaLabel,
-  className,
+  classes,
   testId,
 }: UISelectProps) {
   return (
-    <div data-select className={className}>
+    <div data-select className={classes}>
       {labelText && <label htmlFor="ui-select">{labelText}</label>}
       <UISelect
         data-testid={testId}
--- a/src/components/ui/Select/Select.tsx
+++ b/src/components/ui/Select/Select.tsx
@@ -1,0 +1,54 @@
+import React from 'react'
+import { Select as UISelect } from '@faststore/ui'
+import type { SelectProps } from '@faststore/ui'
+import { CaretDown as CaretDownIcon } from 'phosphor-react'
+
+import './select.scss'
+
+type SelectOptions = {
+  [key: string]: string
+}
+
+interface UISelectProps extends SelectProps {
+  /*
+   * Defines the options available in the select. The SelectOptions object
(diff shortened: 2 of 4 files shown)


src/components/search/Sort/Sort.tsx
```diff
@@ -25,23 +22,15 @@ function Sort() {
...
-      </UISelect>
-      <CaretDownIcon size={18} weight="bold" />
-    </div>
+    <Select
+      className="sort / title-small"
+      labelText="Sort by"
+      options={OptionsMap}
+      onChange={(e) => setSort(keys[e.target.selectedIndex])}
+      value={sort}
+      aria-label="Product Sort"
```
filipewl, line 31: I wonder if we can remove this `aria-label`. The `select`'s accessible name is already being set by its `label`, so there should be no need to use `aria-label` here. It's overriding something that is already working. Any thoughts @hellofanny?
🤔 

CleanShot 2022-02-02 at 14 15 06

src/components/ui/Select/select.scss
```diff
@@ -1,6 +1,6 @@
 @import "../../../styles/scaffold.scss";
 
-.sort {
+[data-ui-select] {
```
filipewl, line 3: A new `data-ui-*` pattern emerges... 👁️

src/components/ui/Select/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+}: UISelectProps) {
+  return (
+    <div data-ui-select className={className}>
+      {labelText && <label htmlFor="ui-select">{labelText}</label>}
+      <UISelect
+        data-testid={testId}
+        onChange={onChange}
+        value={value}
+        aria-label={ariaLabel}
+        id="ui-select"
```
filipewl, line 43: Having this ID hardcoded here is going to be a problem if multiple instances of this component are rendered on the same page because we would have non-unique IDs for different `select`s.

Maybe we should allow an `id` prop so different IDs can be specified?

filipewl (commented):
The extraction is looking good! I left some questions for us to think about.

src/components/search/Sort/Sort.tsx
heitorado: @filipewl the thing is that the label is optional. If we do not provide a text to it, then the aria label is required. I didn't want to overcomplicate things by making the attributes work like a "xor" (only one of them must be provided, but not both).

Since the aria label does not impact on anything 'visual' in the UI, I thought that making it always present regardless of label would maybe simplify things.

src/components/ui/Select/select.scss
heitorado: Oops, that was not my intention. I was just using the component name and was also fearing that it could conflict with some existing `data-select`. But just checked on `faststore` that it is `data-store-select`. 

I'll change it to just `data-select`.

src/components/ui/Select/Select.tsx
heitorado: Hmm, not sure. This `id` sole purpose is to link it with the `label`, right? I think there's not much room for modification there.

Styling can be done by the `data-attributes`, and classes can be added do the wrapping `div`. Do you think of any scenario/use case that adding this "flexibility" to add a custom `id` might be useful?

filipewl: > This id sole purpose is to link it with the label, right? 

Right! 

I'm talking about the scenario of someone using this component multiple times on the same, e.g. to render different `select`s. They would all be rendered with the same `id="ui-select"` attribute, which is a problem because no `id`s should be repeated. This would cause a bug such as clicking on the 2nd `select`'s `label` activating the 1st `select`.
https://codepen.io/voahhsllbutvrxyoje/pen/podbJGa
CleanShot 2022-02-02 at 15 32 33

src/components/search/Sort/Sort.tsx
filipewl: I see, but in this case, we are already providing the `label` via `labelText="Sort by"` so I don't think we need provide `aria-label="Product Sort"`.

hellofanny: I agree with with Filipe! Because we're duplicating unnecessary information. So it might be confusing.

src/components/ui/Select/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+  labelText?: string
+}
+
+export default function Select({
+  options,
+  onChange,
+  labelText,
+  value,
+  'aria-label': ariaLabel,
+  className,
```
hellofanny, line 32:
```suggestion
  classes,
```

hellofanny: Just to make clear that are the extra classes that this component might receive. 
Can you also add this on the interface as optional?

`classes?: string`

src/components/ui/Select/Select.tsx
hellofanny: Please consider what Filipe said. We faced this problem with the `Accodion`. I think for now the best approach is allow an `id` or could be an id prefixe.

src/components/ui/Select/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+import './select.scss'
+
+type SelectOptions = {
+  [key: string]: string
+}
+
+interface UISelectProps extends SelectProps {
+  /*
+   * Defines the options available in the select. The SelectOptions object
+   * keys are the property names, while the values correspond  to the text that
```
hellofanny, line 15:
```suggestion
   * keys are the property names, while the values correspond to the text that
```

src/components/search/Sort/Sort.tsx
```diff
@@ -25,23 +22,15 @@ function Sort() {
...
-        {keys.map((key) => (
-          <option key={key} value={key}>
-            {OptionsMap[key]}
-          </option>
-        ))}
-      </UISelect>
-      <CaretDownIcon size={18} weight="bold" />
-    </div>
+    <Select
+      className="sort / title-small"
```
hellofanny, line 26:
```suggestion
      classes="sort / title-small"
```

renatamottam (approved):
Nice improvement! 😍

src/components/search/Sort/Sort.tsx
heitorado: Oh, now I got it. You're not talking about the `Select` itself, but the implementation done by `Sort`, right? So if I got it right, this issue would be solved by simply not passing an `aria-label` to this implementation of `Select`?

We'll leave the responsibility of always including a `label` XOR an `aria-label` on the user's account? If we want to, I believe we can enforce this logic by using [discriminated union types](https://thoughtbot.com/blog/the-case-for-discriminated-union-types-with-typescript), but not sure if the added complexity is worth it?

src/components/ui/Select/Select.tsx
heitorado: Makes total sense. Thanks people, I'll work on it.

src/components/search/Sort/Sort.tsx
filipewl: Yeah, I think we should remove the prop `aria-label="Product Sort"` in this `Sort.tsx` file.

> We'll leave the responsibility of always including a label XOR an aria-label on the user's account? If we want to, I believe we can enforce this logic by using [discriminated union types](https://thoughtbot.com/blog/the-case-for-discriminated-union-types-with-typescript), but not sure if the added complexity is worth it?

I think it's fine leaving that choice to the user. The logic to enforce the XOR behavior seems like it would overcomplicate `components/ui/Select/Select.tsx`'s implementation.

heitorado: Done at https://github.com/vtex-sites/base.store/pull/299/commits/25065dd8c0c7b682e550839edef64c5af6d21f63

src/components/ui/Select/Select.tsx
heitorado: Done at https://github.com/vtex-sites/base.store/pull/299/commits/40a8fd4988be2ddedd3e93d4502e27f743d2f9ed

saranicoly (approved):
Great job!

filipewl (approved):
Well done!
👏

src/components/ui/Select/Select.tsx
igorbrasileiro: Hey, what's the problem with this hardcoded id? I have no context about it.

src/components/ui/Select/Select.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import { Select as UISelect } from '@faststore/ui'
+import type { SelectProps } from '@faststore/ui'
+import { CaretDown as CaretDownIcon } from 'phosphor-react'
+
+import './select.scss'
+
+type SelectOptions = {
+  [key: string]: string
+}
```
igorbrasileiro, line 10: Record<string, string> has the same meaning of this type, afaik.

src/components/ui/Select/Select.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface UISelectProps extends SelectProps {
+  /*
+   * Redefines the id property to be required when using the Select component. The
+   * id will be used to link the UISelect component and its label.
+   */
+  id: string
+  /*
+   * Defines the CSS class string that will be forwarded to the wrapping div "className" prop.
+   */
+  classes?: string
```
igorbrasileiro, line 21: Why not use the same pattern that react uses, className? By doing this, is transparent for the Select user?

src/components/ui/Select/Select.tsx
```diff
@@ -0,0 +1,64 @@
...
+  /*
+   * Defines the options available in the select. The SelectOptions object
+   * keys are the property names, while the values correspond to the text that
+   * will be displayed in the UI
+   */
+  options: SelectOptions
```
igorbrasileiro, line 27:
```suggestion
  options: Record<string, string>
```

src/components/ui/Select/Select.tsx
```diff
@@ -0,0 +1,64 @@
...
+  /*
+   * Specifies the text that will be displayed in the label right next to the Select.
+   * If omitted, the label will not be rendered.
+   */
+  labelText?: string
```
igorbrasileiro, line 32: What do you think about naming this variable as label?!

src/components/ui/Select/Select.tsx
heitorado: Good one. Done at https://github.com/vtex-sites/base.store/pull/299/commits/24470749d588f91ae1c0b6151c992c4ff3f2c5fa

src/components/ui/Select/Select.tsx
heitorado: You're right. Done at https://github.com/vtex-sites/base.store/pull/299/commits/24470749d588f91ae1c0b6151c992c4ff3f2c5fa

src/components/ui/Select/Select.tsx
heitorado: I spoke with other teammates and we're sticking to this `classes` naming pattern to indicate that this is not the "native" `className` react attribute, but instead a new attribute that comes from the outside and will be forwarded to some react node internal to that component.

src/components/ui/Select/Select.tsx
heitorado: Hmm I'd rather stick with `labelText` to make it explicit in the name that is only the text of the label, not the HTML label itself. To me it makes explicit that the only customizable thing here is the text inside the label.

src/components/ui/Select/Select.tsx
heitorado: Update: Apparently that's only when the interface defined does not extend an interface that already has `className` implemented. You're right, my bad.

src/components/ui/Select/Select.tsx
igorbrasileiro: Yeah, you're right. good point

PR #304, opened 2022-02-04
[FSSS-172] Fix/Adjust inappropriate rerenders

ArthurTriis1:

This PR fixes some inappropriate rerenders with memo function and removing  peer dependencies from hooks. To observe this rerenders I used the [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi) with the option "Highlight updates when components render" checked.

Screen Shot 2022-02-04 at 11 05 10

||Before|After|
|-|-|-|
| PLP | plp-old | plp |
| HOME |  home-old | home |
| PDP | pdp-old | pdp |

### DIFF
--- a/src/pages/{StoreCollection.slug}.tsx
+++ b/src/pages/{StoreCollection.slug}.tsx
@@ -1,21 +1,16 @@
-import { parseSearchState, SearchProvider, useSession } from '@faststore/sdk'
+import { SearchProvider, useSession } from '@faststore/sdk'
 import { graphql } from 'gatsby'
 import { BreadcrumbJsonLd, GatsbySeo } from 'gatsby-plugin-next-seo'
-import React, { useMemo } from 'react'
+import React, { memo, useMemo } from 'react'
 import loadable from '@loadable/component'
+import Hero from 'src/components/sections/Hero'
 import { ITEMS_PER_PAGE } from 'src/constants'
 import { applySearchState } from 'src/sdk/search/state'
 import { Headphones as HeadphonesIcon } from 'phosphor-react'
-import Breadcrumb from 'src/components/ui/Breadcrumb'
-import Hero from 'src/components/sections/Hero'
+import { BreadcrumbWrapper } from 'src/components/ui/Breadcrumb'
+import type { Props } from 'src/hooks/useSearchParams'
+import { useSearchParams } from 'src/hooks/useSearchParams'
 import ProductListing from 'src/components/sections/ProductListing'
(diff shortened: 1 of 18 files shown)


src/pages/{StoreProduct.slug}/p.tsx
```diff
@@ -99,7 +95,7 @@ function Page(props: Props) {
 
       <ProductDetails product={product} />
 
-      {haveYouMightAlsoLikeProducts && (
+      {youMightAlsoLikeProducts?.length > 0 && (
```
filipewl, line 98: Nice clean-up on this file and the others!

src/pages/{StoreCollection.slug}.tsx
```diff
@@ -76,9 +43,6 @@ function Page(props: Props) {
     [allStoreProduct]
   )
 
-  const haveYouMightAlsoLikeProducts =
```
filipewl, line 79: Do you think we can also remove the `youMightAlsoLikeProducts = useMemo` on the lines above? Saw we did that on the PDP!

src/pages/{StoreCollection.slug}.tsx
```diff
@@ -1,4 +1,5 @@
-import { parseSearchState, SearchProvider, useSession } from '@faststore/sdk'
+/* eslint-disable @typescript-eslint/no-use-before-define */
```
filipewl, line 1: Is this still needed? I removed it locally but got no ESLint warning, not even when trying to commit and `eslint --fix` was run.

filipewl (approved):
LGTM! 

Memo all the things

hellofanny:
VERYY NICEE! 🥳 🥳  Well done!! 🚀

src/pages/{StoreCollection.slug}.tsx
ArthurTriis1: I forgot to remove, hehe

filipewl:
I also didn't know about that option for the React extension! **T**oday **I** **L**earned.

eduardoformiga (approved):
👏 👏 👏  Heyyy, Amazing changes! 
Very nice extractions into new hooks, good use of destructuring, etc.
The code looks more legible and better now! 🤩 

I wonder if should we memo small components since maybe keeping the react algorithm rerender can be better for performance in this case. It's more recommendable memo medium to large components.
ref: https://dmitripavlutin.com/use-react-memo-wisely/#2-when-to-use-reactmemo

PR #308, opened 2022-02-07
[FSSS-124] Fix sentinel

heitorado:
This PR revisits `Sentinel` implementation to enhance/fix some of its behavior.

## How does it work?
- It modifies the `Sentinel` to wrap around a product list page by making it accept other components as `children`. This way we can have a better control of page switching when scrolling up and down.
- It adds an `useEffect` hook to handle an edge case of pagination: When we click to load more items but there are no products (and therefore no sentinels) on screen - This was making the pagination behave weirdly.
- [WIP] It enhances the `plp` tests by getting rid of the forced clicks and offsets, that were only a workaround for the sentinel's limitations.

## How to test it?
Go to the PLP, scroll up and down, load products, play a little.
Run `yarn test`, verify that everything passes.

**[Bug Found]**: When scrolling down the PLP, stop right when the pagination changes. Now start scrolling up back to the page you were. Note that the page _**does not**_ change. This happens because the hook only updates pagination when the `inView` for a `Sentinel` changes. On this scenario, neither the upper nor the bottom sentinel had changes on their view status, so pagination is not updated. Not sure how to solve this yet.

## References
Big praise and thanks to @hellofanny

### DIFF
--- a/cypress/integration/plp.test.js
+++ b/cypress/integration/plp.test.js
@@ -144,8 +144,8 @@ describe('Infinite Scroll pagination', () => {
 
         cy.getById('show-more')
           .should('exist')
-          .scrollIntoView({ offset: { top: 50 }, duration: 100 })
-          .click({ force: true })
+          .scrollIntoView({ duration: 500 })
+          .click()
           .then(() => {
             // Ensure it waits for the new page after clicking "show more"
             cy.location('search').should('match', /page=1$/)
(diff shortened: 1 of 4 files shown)


tlgimenes:
This introduces a wild bug where I can see the first page but the url shows `?page=1` beacause the sentinel now has length
image

filipewl:
I'm going to change the target branch to `master` because we're planning to delete the `develop` branch.

filipewl:
I'm going to do a `git rebase --onto origin/master ebaed1bd5785d~1` and a forced push because changing the target branch introduced some dirty.

image

PR #310, opened 2022-02-08
chore: Simplify pages redirects and fixes warning

tlgimenes:
We were doing two kinds of redirects. The first one was a pages redirects, a.k.a /account -> storeframework.myvtex.com/account and the second one was an api redirect, a.k.a /_v/private/graphql to storeframework.myvtex.com/_v/private/graphql. 

The API redirect kind makes no sense for me, since these requests should ALWAYS be made on the myvtex.com domain, so the main store's domain should never receive any request for these apis, so I've removed them

Also, I think page redirects type can be done in a better way. Redirects are platform/framework dependent and are not easy to port to other frameworks/platforms. To avoid this, we can do a react-based redirect on the browser, with the benefit of improving SEO by adding the right tags to these pages, which is not usually the case on other VTEX's services. For instance, VTEX does not add the right no-index/no-follow tags to the `/checkout` page in checkout-ui v6.
 
One nice side effect is that the whole page structure is now located inside the `/pages` folder

Also, this PR fixes the warning:
```
warn There are routes that match both page and redirect. Pages take precendence over redirects so the redirect will not work:
 - page: "/account/" and redirect: "/account/" -> "https://secure.vtexfaststore.com/api/io/account"
 ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Replaces page type redirects, a.k.a. `/account`, `/login` to a corresponding file in `/pages` folder
 
 ### Deprecated
 
 ### Removed
+- API style redirects from `/_v/private/graphql` since they have no effect
 
 ### Fixed
 
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -2,10 +2,6 @@ const path = require('path')
 
 const fs = require('fs-extra')
 
-const config = require('./store.config')
-
-const { secureSubdomain, loginUrl, accountUrl } = config
-
 exports.onPreInit = ({ reporter }) => {
   reporter.info('Copying Partytown Files')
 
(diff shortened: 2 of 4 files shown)


lariciamota (approved):

emersonlaurentino:
This approach has a problem because the status of this redirect is 200, so the search index like Google, doesn't understand what the page should index. It works well, actually redirects, but we have this problem.

tlgimenes:
> This approach has a problem because the status of this redirect is 200, so the search index like Google, doesn't understand what the page should index. It works well, actually redirects, but we have this problem.

that's why we add the noindex nofollow tags on the page

emersonlaurentino:
> > This approach has a problem because the status of this redirect is 200, so the search index like Google, doesn't understand what the page should index. It works well, actually redirects, but we have this problem.
> 
> that's why we add the noindex nofollow tags on the page

So LGTM, once this is allowed on Google.

> Only use JavaScript redirects if you can't do server side or meta refresh redirects. While Google attempts to render every URL Googlebot crawled, rendering may fail for various reasons. This means that if you set a JavaScript redirect, Google might never see it if rendering of the content failed.

emersonlaurentino (approved):

PR #314, opened 2022-02-08
fix: bad react key

tlgimenes:
Fixes the following warning:
image

Also, this PR fixes some bad usages of the Suspense API. Suspense API should NEVER be used during SSR on React 17.x. 

Also, fixes this lighthouse warning on the process
image

### DIFF
--- a/src/components/sections/ProductGallery/ProductGallery.tsx
+++ b/src/components/sections/ProductGallery/ProductGallery.tsx
@@ -1,22 +1,21 @@
+import './product-gallery.scss'
+
 import { usePagination, useSearch } from '@faststore/sdk'
 import { GatsbySeo } from 'gatsby-plugin-next-seo'
-import React, { useState, lazy, Suspense } from 'react'
-import Button, { LinkButton } from 'src/components/ui/Button'
-import Sort from 'src/components/search/Sort'
-import useWindowDimensions from 'src/hooks/useWindowDimensions'
 import {
-  FadersHorizontal as FadersHorizontalIcon,
   ArrowLeft as ArrowLeftIcon,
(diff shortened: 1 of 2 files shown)


emersonlaurentino (approved):

src/components/sections/ProductGallery/ProductGallery.tsx
```diff
@@ -1,23 +1,22 @@
+import './product-gallery.scss'
```
eduardoformiga, line 1: Are there some gains on importing the style file first? If so we can replicate it in the others components.

eduardoformiga (approved):
Nice! 👏 
Should we wait until `Gatsby 4`/ `React 18` to implement the Suspense API?

src/components/sections/ProductGallery/ProductGallery.tsx
tlgimenes: ops, not really, This is just my code editor sorting imports.

tlgimenes:
> Nice! 👏
> Should we wait until `Gatsby 4`/ `React 18` to implement the Suspense API?

I think so. Suspense API can be tricky to use and is not a silver bullet. The more you lazy render the less TBT you have, however, too much lazy/suspense can lead to a TTI and SI increase. The best approach is concentrating them in a single file and use it around the project. That's the reason to be for the `sections` folder

igorbrasileiro (approved):

src/components/sections/ProductGallery/ProductGallery.tsx
tlgimenes: I'm reverting this change. Also, I'm resolving this comment

PR #321, opened 2022-02-08
[FSSS-181] Chore: Moves filter logic to api

eduardoformiga:
Initially, this PR aims to remove the Accordion Item from the Filter component when we have the slug item (disabled) as an option. I mean, when we are on the PLP page. 

When we are on the Search page, the Item with the slugs (usually the department) should appear.

But after [this suggestion](https://github.com/vtex-sites/base.store/pull/321#pullrequestreview-877461915) 

> Please move this processing to the @faststore/api. It will make less processing on the frontend and traffic less data

Then, I created [this PR](https://github.com/vtex/faststore/pull/1154) and remove some logic from the frontend here.

Depends on https://github.com/vtex/faststore/pull/1154

||Before|After|
|-|-|-|
|PLP|Screen Shot 2022-02-21 at 20 35 38|Screen Shot 2022-02-21 at 20 41 49|
|Search Page|Screen Shot 2022-02-21 at 20 39 10|Screen Shot 2022-02-21 at 20 43 05|

## Checklist

- [x] CHANGELOG entry added


danzanzini (approved):

tlgimenes (changes requested):
Please move this processing to the @faststore/api. It will make less processing on the frontend and traffic less data

igorbrasileiro (approved):

tlgimenes (approved):
Nice! Just make sure all checks are passing, but, overall, LGTM 🎉

ArthurTriis1 (approved):

src/components/sections/ProductGallery/useOrderedFacets.ts
```diff
@@ -9,15 +9,7 @@ export const useOrderedFacets = (data?: ProductGalleryQueryQuery) => {
 
   return useMemo(() => {
```
filipewl, line 10: Will we still need this whole file considering the ordering has been moved to the server?

eduardoformiga: I think this file is important for the sake of avoiding the Filter component blinking after rerendering. But considering we now have the skeletons, let me test it again to double-check it.

eduardoformiga: Yeah, We need this `hook` to avoid reloading all the `Filter` and its `Skeleton` every time the user toggle some facet.
But I think we can remove the `useTotalCount` and change a little the `loading` prop criteria from the Sort/totalCount. I'm going to deal with it on another PR.

PR #324, opened 2022-02-09
chore: Remove some procesing from plp

tlgimenes:
PLP was doing one unecessary thing. 

The first one was sorting the facets. This is bad since we need to respect what's coming from the backend since searches may have features of sorting/disabling facets and if we sort on the frontend it may generate a bugs and thus a ticket.

The second weird thing we were doing is creating a hook for accessing the totalCount attribute in a variable. This creates one unnecessary function call and creates an unnecessary ref. 

## How to test it?
Nothing should have changed

## Checklistg
- [x] CHANGELOG entry added


eduardoformiga (commented):
I think we can remove the `useTotalCount`, but then we will have the `0 results` all the time during the loading.
If we remove the `useOrderedFacets` the filter will be rerendered every time, like:
|Previous Version|Changes Introduced|
|-|-|
|NmCAJv5ZtD|NV2gZWttwB|


IMO the previous version with a fixed Filter looks better for the User experience. What do you think?

igorbrasileiro (approved):

filipewl (commented):
@eduardoformiga made some good points. Is there a way to get around those points in this new version?

PR #325, opened 2022-02-09
[FSSS-176] Refactors minicart

danzanzini:
This PR solves https://vtex-dev.atlassian.net/browse/FSSS-176 and https://vtex-dev.atlassian.net/browse/FSSS-183.

## How does it work?
I've followed this to do the refactor:
- Removed <article> in favor of <Card> on CartItem.
- Removed <section> in favor of <List> for the list of Cart Items
- Favored `data-attributes` and tags instead of classes
- Removed the fixed height of the footer.
- Removed the footer of a class named body. 

## How to test it?

Go to the preview, the cart sidebar should work the same way as before.

## Checklist

- [x] CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+
 - Replaces page type redirects, a.k.a. `/account`, `/login` to a corresponding file in `/pages` folder
 - Replaces `let` declarations for `useRef` for better React compatibility
+- Refactors cart sidebar
 
 ### Deprecated
 
 ### Removed
+
 - Unnecessary map at hooks
 - API style redirects from `/_v/private/graphql` since they have no effect
 
--- a/src/components/cart/CartItem/CartItem.tsx
+++ b/src/components/cart/CartItem/CartItem.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { XCircle as XCircleIcon } from 'phosphor-react'
+import { Card } from '@faststore/ui'
 import Button from 'src/components/ui/Button'
 import { Image } from 'src/components/ui/Image'
 import Price from 'src/components/ui/Price'
@@ -20,7 +21,7 @@ function CartItem({ item }: Props) {
   const { updateItemQuantity } = useCart()
 
   return (
-    <article
+    <Card
       className="cart-item"
       data-cart-item
       data-testid="cart-item"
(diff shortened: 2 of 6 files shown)


src/components/cart/CartItem/CartItem.tsx
```diff
@@ -20,7 +21,7 @@ function CartItem({ item }: Props) {
   const { updateItemQuantity } = useCart()
 
   return (
-    <article
+    <Card
```
hellofanny, line 24: Why didn't you use the other card elements (`CardContent`, `CardImage`, `CardActions` ...) to compound this `CartItem`?

danzanzini: Didn't know about them. Will look at it. Thanks

hellofanny: okay! I just asked, but you don't have to use it if not seems necessary :)

renatamottam (approved):
Great work, dan! You're nailing CSS 😍

CHANGELOG.md
```diff
@@ -10,12 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+
 - Replaces page type redirects, a.k.a. `/account`, `/login` to a corresponding file in `/pages` folder
 - Replaces `let` declarations for `useRef` for better React compatibility
+- Refactors cart sidebar
```
hellofanny, line 16: Maybe we can give more details here, what do you think?

```suggestion
- Refactors cart sidebar using `Card` and `List` from `@faststore/ui`
```

src/components/cart/CartItem/CartItem.tsx
```diff
@@ -20,30 +21,31 @@ function CartItem({ item }: Props) {
...
+            backgroundColor="#f0f0f0"
+            options={{
+              fitIn: true,
+            }}
+          />
+        </CardImage>
         <div data-cart-item-summary>
           <p className="text-body">{item.isVariantOf?.name}</p>
-          <div data-cart-item-price>
+          <span data-cart-item-price>
```
hellofanny, line 48:
```suggestion
          <span data-cart-item-prices>
```

Just because it wrappers more than one price 👀

hellofanny: If you agree with this suggestion, don't forget to replace it on the css too, ok? 😉

saranicoly (approved):
Great job, Dan!! 👏🏾👏🏾

CHANGELOG.md
danzanzini: The problem is that it was more than this. I've also worked to remove the fixed height, to change how the HTML was organized, and to remove some classes. Don't know how to wrap those on the changelog.

PR #340, opened 2022-02-16
refactor: Remove useWidowDimensions hook

tlgimenes:
This PR removes the useWindowDimensions hook.

## How does it work?
React rendering is based on a concept called VirtualDOM. This means that, before rendering anything to the DOM (creating divs, anchors, etc), React renders a representation of the App's state into this VirtualDOM, applying side effects and waiting for the App's state to stabilize. After the VirtualDOM is stable, React goes into a "commit" phase, where it renders the difference between the VirtualDOM state into the DOM, triggering the browser's rendering pipeline.
This VirtualDOM approach is a good performance advantage of React because it saves the browser from rendering ephemeral, intermediate states, and makes the browser focus only on useful states of your app. This approach, however, has some peculiarities and constraints that must be strictly respected when dealing with Server Side Rendering (SSR) 

Server side rendering is the technique of running React on the server, pre generating html tags that are them written to the site's final html file. This drastically improves FCP/LCP, since the browser can render and fetch resources much earlier than a pure client-side (CSR) app would. Below you can see the difference in the rendering pipeline of a pure CSR vs SSR App. 

CSR: 
1. Request html to the server
2. The html is downloaded
3. Downloads external resources (JS, CSS)
4. React mounts and starts rendering
5. App is rendered (FCP)
6. Images are downloaded (LCP)

SSR: 
1. Request html to the server
2. The html is downloaded (FCP)
3. Downloads external resources (JS, CSS, Images) (LCP)
4. React mounts and hydrates

As you can see, FCP and LCP happens much earlier on a Server Side Rendered App. This is because the Server already processed all the necessary DOM nodes required to represent(render) that web page before the client's browser even requested anything, leaving the browser with the only task of rendering the whole page at once.
This works perfectly for static, Javascript-less application, like a blog etc. React, however, was designed to create interactive web pages, like facebook, where JavaScript is essential for the user interaction. A conciliation, then, is needed between the DOM nodes generated on the Server and the VirtualDOM react generates on the browser. This conciliation process is called hydration

Hydration usually occurs after the browser's FCP (after the user is seeing something on the screen), and thus, after the initial DOM was constructed by the browser. Since React needs a VirtualDOM to operate, React renders the whole app, creating the necessary VirtualDOM. From there, two things may happen:
1. The DOM and VirtualDOM match perfectly
This means that the DOM have the exact same nodes as the VirtualDOM (divs, anchors etc) and now React assumes it has a correct representation of DOM nodes on VirtualDOM and React can start behaving normally.

2. The DOM and VirtualDOM don't match
If the DOMs do not match perfectly, React favors the newly created VirtualDOM, replacing all SSRed DOM nodes with CSR ones. This is called a `hydration mismatch` and may cause the App to blink, triggering the whole rendering pipeline again. 
Also, this may cause a bad user experience and a big performance hit. 

How do I avoid hydration mismatches?
As a rule of thumb, the first render on the browser of a SSR app must match exactly what's been rendered on the server.

### Examples
A common occurrence of hydration mismatches is when using the localstorage. Suppose we have to create a login component that prints `login` if the user is anonymous, and `hi {user.name}` if the user is logged in. Also, suppose we will store the user's data on the localstorage. One may create the following code:

```tsx
import React, { useState } from 'react'

const useProfile = () => {
  const [user, setUser] = useState<{ name: string } | null>(
    typeof window !== 'undefined' ? JSON.parse(window.localStorage.getItem('user')) : null
  )

  return user || null
}

function Greeting() {
  const user = useProfile()

  if (user) {
    return <div>Hi {user.name}</div>
  }

  return <div>login</div>
}
```

This works fine if the user has ever logged in, however, this breaks the hydration on the device of recurring users. This is because the server will render `<div>login</div>`, while the first render on the client side will render `<div>hi Foo</div>`. Since the DOM nodes are different, React may detect a hydration mismatch and the app may either blink or enter an inconsistent state. 

How do we solve this? To solve this problem we can change the `useProfile` hook to perform a "two pass" render. The first render will match everything rendered on the server, while the second will render the data on the user's device. This way we ensure the hydration process happens correctly and the user sees their name on the greeting message.
```tsx
import React, { useState, useEffect } from 'react'

const useProfile = () => {
  const [user, setUser] = useState<{ name: string } | null>(null)

  useEffect(() => {
    setUser(JSON.parse(window.localStorage.getItem('user')))
  }, [])

  return user || null
}
```

### Problems with `useWindowDimensions`
The problem with the `useWindowDimensions` hook is that it changes the initial value of a state variable depending on whether React is running on a big or mobile device. This leads to hydration mismatches or bugs. More specifically, this hook was generating an inconsistent state on the footer, where the desktop version was being displayed on mobile devices, breaking the user experience. 
This bug was not noticed before because the footer was being wrapped in a `loadable` function that made the component's rendering occur on client-side only, not displaying the problem. After #328, however, the footer is being rendered on the server, leading to inconsistencies.
image

To fix the problem, I created two classes, `display-mobile` and `hidden-mobile`. These classes use media query to add a `display: none` property and hide, via css, a section of the page.

This solution is far from ideal because ships unnecessary code to mobile/desktop devices. This, however, is the only solution Jamstack offers for dealing with responsive layouts and the designers should take this into account when designing components so that they are responsive for all screen sizes and we don't need different components for different devices.

## How to test it?
Everything should work as before.

## References
To know more about hydration: 

https://www.gatsbyjs.com/docs/conceptual/react-hydration/
https://reactjs.org/docs/react-dom.html#hydrate

## Checklist
- [] CHANGELOG entry added


sergiofontes:
What a great description of the issue, Gimenes; thank you for that! Ideally, we shouldn't rely on JS to trigger styles for a specific device. Instead, we must let the CSS handle this situation.

One caveat that I found is that we should avoid using `display: none` on structures with meaningful content/semantics for the user. That's because the content under a `display: none` flag becomes inaccessible for assistive technologies, like a screen reader. Use the following snippet to visually hide a text (or an area) while still keeping its content accessible:
```
position: absolute;
padding: 0;
margin: 0;
overflow: hidden;
line-height: 0;
border: 0;
clip-path: inset(0 0 0 0);
```

eduardoformiga:
Good point, @sergiofontes.
For meaningful content/semantics, we also have the [SROnly](https://github.com/vtex-sites/base.store/blob/master/src/components/ui/SROnly/SROnly.tsx) component that has a [similar style](https://github.com/vtex-sites/base.store/blob/master/src/components/ui/SROnly/sr-only.scss) that you mention above.

For the case from this PR, IMO we can use the `display: none` toggle since when we hide some component, another similar component is shown elsewhere. So the page remains accessible. What do you think?

eduardoformiga (approved):
Good Job! 👏 

Not related to changes on this PR. But as we are touching the `Filter` here what do you think about refactoring the [onFacetChange](https://github.com/vtex-sites/base.store/pull/340/files#diff-c279b43887308802637d360aa322485ff0b38b30e4c2ff7c542ff72233b35c1cR114-R130) function to the suggestion below. So that we can avoid 2 `.some` loops:

```javascript
const onFacetChange = (item: IStoreSelectedFacet) => {
    const indexToRemove = selectedFacets.findIndex(
      (f) => f.value === item.value
    )

    const shouldRemoveItem = indexToRemove !== -1

    if (shouldRemoveItem) {
      setFacetsToRemove([...facetsToRemove, item])
      selectedFacets.splice(indexToRemove, 1)
      setSelectedFacets([...selectedFacets])

      return
    }

    setSelectedFacets([...selectedFacets, item])
  }
```

eduardoformiga:
Great! Let's do it :)

filipewl (approved):
Thanks for the thorough explanation!
🙇

igorbrasileiro (approved):
@tlgimenes can you add a CHANGELOG entry!? Approved, but add the changelog 😅

PR #329, opened 2022-02-10
fix: Use Gatsby-Link on Link component

igorbrasileiro:
## What's the purpose of this pull request?
Use GatsbyLink component to do client-side navigation.

## How does it work?
Pass to  `GatsbyLink` component to `as` prop from @faststore/ui `Link`.

## How to test it?
Click on store links and check if the navigation is client-side.

## Checklist
- [x] CHANGELOG entry added

PR #344, opened 2022-02-16
fix: Scroll lock when transitioning on mobile via slide-over nav

filipewl:

After the changes from https://github.com/vtex-sites/base.store/pull/329, a bug was accidentally introduced when navigating using one of the links inside the mobile slide-over navigation: scrolling was locked because the `.no-scroll` CSS class was kept in the `#layout` element. The `.no-scroll` is added when the slide-over is opened to prevent scrolling the content on the background. This happened because the links have been replaced by Gatsby's so there isn't a full page refresh anymore.

## How does it work?

The fix works by removing the `.no-scroll` CSS class when needed.

[Before](https://sfj-2ce7b26--base.preview.vtex.app/)|[After](https://preview-344--base.preview.vtex.app/)
-|-
https://user-images.githubusercontent.com/381395/154371235-2d34df1c-4c51-47b1-9fda-5780b74ce4af.mp4|https://user-images.githubusercontent.com/381395/154371303-e37a4f7d-8689-46c4-a508-7452c895f791.mp4

## How to test it?

https://preview-344--base.preview.vtex.app/

1. On mobile viewport, visit a page like the [Home](https://preview-344--base.preview.vtex.app/).
2. Open the slide-over navigation and click on any of the collection/PLP/categories links.
3. Upon arriving on the new page, you should be able to scroll freely.

## References

- https://vtex-dev.atlassian.net/browse/FSSS-201


eduardoformiga (approved):
LGTM! 👏 🎉

src/components/ui/SlideOver/SlideOver.tsx
```diff
@@ -56,6 +56,10 @@ const SlideOver = ({
         layout.current.classList.add('no-scroll')
       }
     }
+
+    return () => {
+      layout.current?.classList?.remove('no-scroll')
```
filipewl, line 61: Now I'm not happy about where I placed this... This component has 3 `useEffect`s, but adding this to the 1st one doesn't fix the issue, while it works for the 2nd and 3rd ones. Thinking where the is best logical place to place this.
🤔 

CleanShot 2022-02-16 at 20 30 46@2x

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - React tree re-rendering
 - Footer rendering pipeline
+- Scroll lock when transitioning pages on mobile via slide-over navigation
```
hellofanny, line 30:
```suggestion
- Scroll lock when transitioning pages on mobile via `SlideOver` component navigation
```

src/components/ui/SlideOver/SlideOver.tsx
hellofanny: Hmm. Actually, I think it's okay to use multiple effects here. 🤔 

As suggested by the documentation: [Use Multiple Effects to Separate Concerns](https://reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns)

hellofanny: Btw, I don't think to verify `handleClose` (line 66). Can you check it again and then remove it pls ✨

hellofanny (approved):
Nice fix! ✨

src/components/ui/SlideOver/SlideOver.tsx
lucasfp13: I also think it's okay to leave on that `useEffect` as it's related to the `layout`.

I believe it didn't work for the first effect because when the cleanup function is called the component is already on unmount state, possibly without the current `layout` value, or possibly it has changed. You can try to store the `layout.current` value inside the `useEffect` scope to keep the value (see this discussion [here](https://stackoverflow.com/questions/67069827/cleanup-ref-issues-in-react))

israelswf (approved):
✨

ArthurTriis1 (approved):

tlgimenes (approved):
nice!

PR #348, opened 2022-02-17
chore: Remove gatsby-plugin-offline

tlgimenes:
This PR removes the `gatsby-plugin-offline` plugin

## How does it work?
Gatsby plugin offline is the recommended way of adding service worker support to Gatsby. It caches important assets, like JavaScript and CSS for faster loading times on recurring users. Also, this plugin implements a pattern called [app-shell pattern](https://developers.google.com/web/fundamentals/architecture/app-shell) that aims to give users the impression of faster loading times by rendering the apps shell before the app's content. 
The way this plugin implements this pattern is by intercepting any reload the user does, returning the app shell and doing a client-side navigation to the desired page. This works well on some projects, but on ours, this pattern was creating a huge layout shift. The left side of the image below shows this LS, where the app's shell is loaded, then a client side navigation is performed and the app's content is loaded. The image on right shows the same device performing the load without this plugin.
<div style="display: flex; justify-content: center">
 

</div>

Also, I noticed that another of solving this problem is by adding a `min-height` to the `<main/>` tag so that this CLS is not visible. 

I think PWA support is not currently necessary and we may go back to this subject once we have fixed other performance issues later on. Also, I noticed that Gatsby's lighthouse score increased after removing this plugin. I'd say we remove this for now and go back to it later on.

> Note that the service worker I'm removing don't really affect partytown or any other pixel scripts

## How to test it?
Make sure you have enabled service workers and cache on devTools. Also, go to the `performance` tab and throttle your CPU 6x. Perform a reload of this PR vs master. You will see the difference

## Checklist
- [x] CHANGELOG entry added


filipewl:
I [had tested](https://github.com/vtex-sites/base.store/pull/306) removing this plugin and noticed a great improvement in Gatsby Cloud's LH score for Performance from [82](https://github.com/vtex-sites/base.store/pull/307#issuecomment-1031906349) to [91](https://github.com/vtex-sites/base.store/pull/306#issuecomment-1031789091), but I couldn't explain why and wasn't sure we could remove it. 

Hoping to learn more about your tests here!

tlgimenes:
> I [had tested](https://github.com/vtex-sites/base.store/pull/306) removing this plugin and noticed a great improvement in Gatsby Cloud's LH score for Performance from [82](https://github.com/vtex-sites/base.store/pull/307#issuecomment-1031906349) to [91](https://github.com/vtex-sites/base.store/pull/306#issuecomment-1031789091), but I couldn't explain why and wasn't sure we could remove it.
> 
> Hoping to learn more about your tests here!

Nice! I didn't know about it. Yeah, I think we can remove it for now and go back to it later on

danzanzini (approved):
🚀 

@tlgimenes Do you know why this was relevant for gatsby's lighthouse and not for ours? It is something that has a visual impact and our lighthouse didn't catch it. I wonder if we should change any config so it will catch behaviors like this in the future.

igorbrasileiro:
You will [need to use another plugin to remove](https://www.gatsbyjs.com/plugins/gatsby-plugin-offline/#remove) the gatsby plugin offline

igorbrasileiro:
> 🚀
> 
> @tlgimenes Do you know why this was relevant for gatsby's lighthouse and not for ours? It is something that has a visual impact and our lighthouse didn't catch it. I wonder if we should change any config so it will catch behaviors like this in the future.

I'm guessing, but I think this difference is related to LHCI version. Gatsby cloud use LH version 7.1, WebOps use version 8.0 and PSI use version 9.0
image

PR #349, opened 2022-02-17
chore: uses absolute path to import stylesheets

hellofanny:

This PR replaces relative stylesheets imports with absolute path and uses the same import file (`scaffold.scss`) on all needed stylesheets files.

## How does it work?

- We were using more the relative path. But with the absolute path It’s easier to use and localize the file.

- Uses `src/styles/scaffold` instead of alternating between it and `vendor/include-media` and `utilities.scss`
     - it helps to maintain a pattern, and we don't need to change the imported file whenever we need to use functions from `utilities.scss` file.
     
## How to test it?

Nothing should have changed. Everything should work fine.

## Checklist

- [x]  CHANGELOG entry added


filipewl (approved):
Nice!

I just wish we could enforce this via a linter, such as Stylelint we use for SCSS, but I wasn't able to find anything about it in my brief research. 😕 We'll just have to policy ourselves from now on!

eduardoformiga (approved):
LGTM! ✨

israelswf:
@hellofanny, just noticed  this `=` in branch's name `chore/FSSS-208-use-absolute=path`, it just grabbed my attention if 
 would cause any problems. But since CI already ran, I think there is nothing to worry about.

hellofanny:
> @hellofanny, just noticed this `=` in branch's name `chore/FSSS-208-use-absolute=path`, it just grabbed my attention if
> would cause any problems. But since CI already ran, I think there is nothing to worry about.

Omo! Well noticed! Thanks!

PR #356, opened 2022-02-21
feat: Remove postal code using query string

igorbrasileiro:
Remove the postal code component from being downloaded and rendered via query string.
⚠️ Please, give me better names for components and hooks.

## How does it work?
When the page URL has `__performanceTest` query string, the PreventLoadComponent prevents the PostalCodeInput from being loaded.

## How to test it?
Add the `__performanceTest` query string from the URL, and check that the browser does not fetch the PostalCode asset.

## Checklist

- [] CHANGELOG entry added


gatsby-browser.js
```diff
@@ -12,6 +12,7 @@ import ErrorBoundary from './src/sdk/error/ErrorBoundary'
 import TestProvider from './src/sdk/tests'
```
tlgimenes, line 12: Hydration provider seems to be duplicated with this one

tlgimenes (commented):
I didn't get the overall intent of this PR.

Postal code component should not impact the performance in any way, so why is it behind a flag called `__performanceTest`?

igorbrasileiro:
> I didn't get the overall intent of this PR.
> 
> Postal code component should not impact the performance in any way, so why is it behind a flag called `__performanceTest`?

Yeah, ok. The postal code now, is just an input, right? But in near future, it will be a modal with more information and more code, of course. This new features may decrease the performance, because of this, I created this PR.

We can hold on, and use it if necessary.

igorbrasileiro:
@tlgimenes Take a look here, https://github.com/vtex-sites/base.store/pull/388/

PR #357, opened 2022-02-21
feat: webpack analyser

tlgimenes:
This PR adds `gatsby-plugin-webpack-bundle-analyser-v2` plugin so we can better visualize what's in the bundle. Also, this PR removes some blobs from the `<Footer/>` component and lazy-load them

## How does it work?
This Gatsby plugin is a wrapper to the famous `webpack-bundle-analyser` plugin. It allows you to visualize what are the main parts of a bundle and how they relate to the total bundle size. 
When you run `gatsby build`, a file called `bundle-analyser.html` will be generated at the `/public` folder. Opening this file reveals the image below. 
image

By analyzing the image, this `Icons.tsx` on the bottom left corner caught my attention. I then realized these svgs where stringified into the code. Stringifiying content makes sense when dealing with above the fold content, like the store's logo, but may bloat the bundle when doing with below the fold content, like on this case. (These icons were being used on the  footer). To solve the problem I moved these icons to `/static/icons` and  lazy loaded them by using an `` tag. 
After the changes, we get the following bundle-analyser report:
image

As you can see, we have no more icons on the store's final bundle. 

To make sure the bundle size had decreased, I used the other tool we use, called `bundle-stats.html`. Below, we can see that the initial bundle size shrunk from 393 to 360Kb in size, a 9% decrease in bundle size.

image
image

Also, a nice side effect was reducing total html size, from 38.2Kb to 21.7Kb, a 23% reduction in html size 🎉 

## How to test it?
Everything should work exactly as before

## Checklist
- [x] CHANGELOG entry added


danzanzini (approved):
🚀 Great job, and thanks (again) for the PR description

Do you think it's worth adding information regarding these tools to our README?

I've also created a `performance` label so we can group all our PRs related to performance and bundle size, what do you think?

hellofanny:
Wow!! Great!
I love to read these PR descriptions! Thank you! 🥇

hellofanny (approved):

filipewl (approved):
Very nice!

igorbrasileiro (approved):

igorbrasileiro:
Something interesting is that the 32 module concatenated from phosphor. Have you looked these modules @tlgimenes ?

tlgimenes:
> I love to read

Not really, but I think it's not really affecting the performance that much, but it's worth testing

PR #317, opened 2022-02-08
[FSSS-108] Feat: Add Skeleton Loading

lucasfp13:
## What's the purpose of this pull request?

This PR intends to add Skeleton Loading components to be visible while the real components are being loaded.

The Skeleton components will be used as a fallback for lazy-loaded components or while data is not ready.

## How does it work?

We can observe the behavior when accessing a page. As long as the data is not ready, the skeleton components will be visible.

_Desktop version:_

Screen Shot 2022-02-09 at 17 25 42
Screen Shot 2022-02-09 at 17 25 55
Screen Shot 2022-02-09 at 17 26 59

_Mobile version:_

Screen Shot 2022-02-10 at 01 06 09Screen Shot 2022-02-10 at 01 05 07
Screen Shot 2022-02-10 at 01 05 21

## References

- Related to offload the animations to the GPU: [Optimizing Repaints](https://www.afasterweb.com/2017/07/27/optimizing-repaints/);

## Checklist

- [x] CHANGELOG entry added

PR #358, opened 2022-02-22
chore: lazy load icons

igorbrasileiro:
This PR removes all phosphor icons from the app bundle. This improves performance and decreases the size of the final bundle. Doesn't remove the icons from CartSideBar because it is already lazy-loaded.

Motivation: Phosphor react icons are really good, but they are heavy and they were on the final app bundle. So lazy loading them will remove them from the final bundle.
As you can see, in red is the phosphor react components
**Before**:
image

**After**:
image

Also, the final document decreased the size from 21.7kb to 20.0kb according to the https://pagespeed.web.dev
Finally, it reduced the app size 
**from**:
image
**to**:
image

⚠️ This PR changes the **lighthouse** settings.

## How does it work?
Part 1:
Created a LazyLoad component that renders the Phosphor icon lazily. The webpack will create new chunks for each lazy icon.
Part 2:
Change the lighthouse assertions, to allow fetch more scripts. This is needed because part 1 creates more scripts for each icon.

## How to test it?
Check the evidence:
https://preview-358--base.preview.vtex.app/bundle-analyser.html
https://base.vtex.app/bundle-analyser.html

Also, compare the preview with the main URL to ensure that's the icons are rendering at the same size and positions.

## References
https://reactjs.org/docs/code-splitting.html

## Checklist
TBD

- [] CHANGELOG entry added


src/components/cart/CartSidebar/CartSidebar.tsx
```diff
@@ -22,6 +18,28 @@ import './cart-sidebar.scss'
...
+const TruckIcon = lazy(() => import('phosphor-react/src/icons/Truck'))
+
+const LazyIcon = ({
+  icon: Icon,
+  size = 32,
+}: PropsWithChildren<{
+  size?: number
+  icon: ElementType<{ size: number }>
+}>) => {
+  if (typeof window === 'undefined') {
```
tlgimenes, line 32: conditional rendering mismatch

renatamottam:
@igorbrasileiro, do you think we could add a skeleton to them?
Reference: https://github.com/vtex-sites/base.store/pull/317

igorbrasileiro:
Sometimes the PSI score is at ~50. Does it is normal, @tlgimenes ?

igorbrasileiro:
Closing this PR, because it's decreasing the performance. If I discover a better way to remove those icons from the final bundle without harming the performance on PSI, I will reopen. Also, I will add the skeleton

src/components/common/IconSVG/IconSVG.tsx
```diff
@@ -0,0 +1,27 @@
+import type { ImgHTMLAttributes } from 'react'
```
tlgimenes, line 1: This is a dumb component. I'd rather move it to /components/ui and wrap it using the `Icon` component from '@faststore/ui' so we don't need this `data-icon` attribute to style the icon, what do you thinnk? Also, it could be named of Icon, instead of IconSVG.

igorbrasileiro:
Closing this PR in favor of this https://github.com/vtex-sites/base.store/pull/378

PR #360, opened 2022-02-22
fix: remove from cart and analytics events

tlgimenes:
This PR solves an issue when removing a specific item from cart. For comparison, check the before/after gif below:

<div style="display: flex">
before
after
</div>

## How does it work?
We were not retrieving all necessary data from the server on cart validation. This caused, that, for some products, these necessary infos were lacking, that, in turn caused the following null pointer exception on the `useRemoveButton` hook:
image

This class of "null pointer exception" errors should be impossible to happen on `base.store` due to the nature of our end-to-end typings powered by GraphQL and TypeScript. Taking a further look, I've figured out that a type cast was the root of the problem. 
The `CartItemWithAnalytics` type, alongside the cast present on `useCart` hook, broke TypeScript typings, making it possible to have objects of incomplete types on the store. To solve the issue, I now removed the type `CartItemWithAnalytics` and added the necessary info to the `validateCart` mutation. 

Also, another problem with this was that we were using the wrong sku id for the product, what caused the wrong product to be added to cart. This PR solves this issue and should be merged before this one
https://github.com/vtex/faststore/pull/1156

## How to test it?
Try to add/remove products from the store and make sure all are removeable, specially this one: `/handcrafted-concrete-soap-36085291/p`

## Checklist
- [x] CHANGELOG entry added


igorbrasileiro (approved):

saranicoly (approved):
LGTM! 👏🏾👏🏾

igorbrasileiro (commented):
Rember tu set a stable version of @faststore/api

filipewl (approved):
LGTM!

We also need to get it merged quickly because FS v1.5.15 has a [fix](https://github.com/vtex/faststore/pull/1157) to our current breaking [builds](https://github.com/vtex-sites/base.store/runs/5293982513).

PR #363, opened 2022-02-23
chore: fix Hero Image sizes and css

igorbrasileiro:
This PR fixes the image HTML sizes property, by doing this you let the browser choose a better image to fetch.
Nowadays the Image component uses the gatsby-image-plugin that sets a default `100vw` for sizes. For small screens with size less than 768px, must fetch images with the 100vw size or higher, but for screens higher than 768px the Hero component renders a image with the half of screen width.

Because of it on devices with bigger screens was fetching an image with high resolution without necessity.
**Before**:
image

**After** 
image

⚠️ Pay attention on images sizes and links that has resolution of image.

Also, remove unnecessary padding that the gatsby-image-plugin set.

## How does it work?
Set 100vw for screens with max-width = 768px and 50vw for higher screen width.

## How to test it?
Open the preview and see the Hero image size versus the same image on production https://base.vtex.app

## References
https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

## Checklist
TBD
- [] CHANGELOG entry added

### DIFF
--- a/src/components/sections/Hero/Hero.tsx
+++ b/src/components/sections/Hero/Hero.tsx
@@ -69,8 +69,13 @@ const Hero = ({
           options={{
             fitIn: true,
           }}
+          // reset gatsby image default style
+          style={{
+            overflow: undefined,
+            position: undefined,
+          }}
           // for mobile load 100vw image, for desktop load half img
-          sizes="(max-width: 768px) 100vw, 50vw"
+          sizes="(max-width: 768px) 100vw, 53vw"
         />
       </HeroImage>
     </UIHero>
--- a/src/components/sections/Hero/Hero.tsx
+++ b/src/components/sections/Hero/Hero.tsx
@@ -69,6 +69,8 @@ const Hero = ({
           options={{
             fitIn: true,
           }}
+          // for mobile load 100vw image, for desktop load half img
+          sizes="(max-width: 768px) 100vw, 50vw"
         />
       </HeroImage>
     </UIHero>
--- a/src/components/ui/Hero/hero.scss
+++ b/src/components/ui/Hero/hero.scss
@@ -59,6 +59,11 @@
   }
 
   [data-hero-image] {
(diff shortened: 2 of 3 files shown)


src/components/ui/Hero/hero.scss
```diff
@@ -59,6 +59,11 @@
   }
 
   [data-hero-image] {
+    // override 50% padding top from component
+    [aria-hidden][style]:first-child {
+      padding-top: 100% !important; /* stylelint-disable-line declaration-no-important */
```
hellofanny, line 64: I haven't noticed any visual difference. So, I didn't get this padding, does it make any other significant change? 😕 
`padding: 50%`
image

`padding: 100%`
image

src/components/sections/Hero/Hero.tsx
```diff
@@ -69,6 +69,13 @@ const Hero = ({
           options={{
             fitIn: true,
           }}
+          // reset gatsby image default style
+          style={{
+            overflow: undefined,
+            position: undefined,
+          }}
+          // for mobile load 100vw image, for desktop load half img
+          sizes="(max-width: 768px) 100vw, 53vw"
```
hellofanny, line 78: We have similar approach on [ProductCard](https://github.com/vtex-sites/base.store/blob/dd8df7195b5191e96f68900d3db1b8ad806245cc/src/components/product/ProductCard/ProductCard.tsx#L31). 

Maybe `50vw`, instead of `53vw`?

igorbrasileiro: The size of the current image is 53% of the whole screen

igorbrasileiro: 53vw will get an image with a higher resolution, that works for this 53% image width and for PDP that uses 50%. Does it make sense?

hellofanny (approved):

PR #367, opened 2022-02-24
feat: adds EmptyState component

danzanzini:

This PR solves [FSSS-196](https://vtex-dev.atlassian.net/browse/FSSS-196). It creates an `EmptyState` component to use at the `CartSidebar` and at the `ProductGallery`.

## How does it work?

This new component basically styles all elements sent to it as children. The reason why I've done with children is that the majority of the component (the actions) already needed to be passed as children, so I didn't see a reason to insert two props for the title and icons.

I couldn't make the buttons smaller as it's stated in Figma because the `height` would be less than the `--tap-min-size` value, so I let it as it is.

## How to test it?
Just go to the preview and test it. Open the cart with no products on it and make a search that brings no results. 
Here are some screenshots:

### New cart empty state:
Screen Shot 2022-02-24 at 7 09 05 PM

### Product Gallery Empty State
ezgif com-gif-maker

## Checklist
- [x] CHANGELOG entry added
- [x] Decide the reference for the two links


israelswf (approved):
LGTM, Dan! 
Just a question, the buttons `Browse Offers` and `Just Arrived` will stay without actions for this task?

src/components/common/EmptyState/empty-state.scss
```diff
@@ -0,0 +1,32 @@
...
+
+.empty-state[data-empty-state] {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  margin-right: rem(101px);
+  margin-left: rem(101px);
+  padding-top: rem(95px);
+  padding-bottom: rem(95px);
```
renatamottam, line 11:
```suggestion
  padding-bottom: var(--space-5);
  padding-top: var(--space-5);
```

renatamottam (approved):
Nice work! 😍 
Just left some improvement suggestions. I would additionally add these paddings:
```
.product-listing {
    // use  var(--space-5) for desktop
    padding-top: var(--space-4);
    padding-bottom: var(--space-4);
} 
```

hellofanny (commented):
Well done Dan! Actually, I prefer this approach that you implemented 🙈 ; it gives us more flexibility to compound our Empty State layouts. Regarding the button's height, no worries at all. We need the buttons to respect the `--tap-min-size` size because of [LH](https://web.dev/tap-targets/), maybe we can update this on Figma to match it (or find another way to approach this issue) 🙃 

Just a note on the mobile version:
- In the prototype, we don't have the borders gap. Not sure if it was decided to keep it 😕 

|Preview|Prototype|
|-|-|
Preview|Prototype

src/components/common/EmptyState/empty-state.scss
```diff
@@ -0,0 +1,37 @@
...
+.empty-state[data-empty-state] {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  min-height: 50vh;
+  margin-top: var(--space-5);
+  margin-bottom: var(--space-5);
+  padding-top: var(--space-5);
+  padding-bottom: var(--space-5);
```
hellofanny, line 12:
```suggestion
  padding: var(--space-5) 0;
```

hellofanny: Maybe, we could just use the shorthand property here 💡

hellofanny:
> Well done Dan! Actually, I prefer this approach that you implemented with `children` over specific `props` 🙈 ; it gives us more flexibility to compound our Empty State layouts. Regarding the button's height, no worries at all. We need the buttons to respect the `--tap-min-size` size because of [LH](https://web.dev/tap-targets/), maybe we can update this on Figma to match it (or find another way to approach this issue) 🙃
> 
> Just a note on the mobile version:
> 
> * In the prototype, we don't have the borders gap. Not sure if it was decided to keep it 😕
> 
> Preview	Prototype
> Preview	Prototype

Renatinha said they decided to add the padding on the mobile version because it was too close to the footer.

hellofanny (approved):

filipewl (approved):
LGTM!

PR #368, opened 2022-03-03
chore: Performance improvements by sections

tlgimenes:
This PR uses the new [content-visibility](https://web.dev/content-visibility/) css property to improve layout loading of the page. 

## How does it work?
We organize our page layouts in `sections`. These sections are self contained strips of the page that offers a feature (carousel, shelf, hero etc).
Turns out, that for such architecture, W3C has recently launched a new CSS feature to `lazy load` part of the page's content so browsers do less work. This feature is called `content-visibility` and better explanation is done in [web.dev's site](https://web.dev/content-visibility)

To improve our sections architecture, I created a new `Section` component that encapsulates this css rules with the `section` selector.

Below, you can see, with 95% confidence interval, rendering times for the home page:

before: 396 +- 18
after: 328 +- 11

which is ~20%  improvement in rendering times of the home page. This is a small improvement, however, given the small effort, I guess we can move forward with this PR

## How to test it?
Nothing should have changed

## Checklist
- [x] CHANGELOG entry added


src/components/sections/Section/Section.scss
```diff
@@ -0,0 +1,9 @@
+.pages__section {
+  width: 100vw;
+  content-visibility: auto;
+  contain-intrinsic-size: 2px 1024px;
```
filipewl, line 4: Why these two values? It [seems](https://drafts.csswg.org/css-sizing-4/#:~:text=The%20first%20value%20represents%20the%20contain%2Dintrinsic%2Dwidth%20value) they're the width and height, respectively. Have you gotten them inverted by any chance (1024px is a common width value)? Considering the responsive versions, maybe we can replace the fixed 1024px for something like 90vw/100vw/% or something.

Could we use `auto` as stated [here](https://web.dev/content-visibility/#:~:text=is%20a%20new-,auto%20keyword,-for%20contain%2Dintrinsic) to help with the [last remembered size](https://drafts.csswg.org/css-sizing-4/#last-remembered)?

Maybe we could take these values with the mobile version as the baseline with the width as 90/100vw and the height of 380px, something like `contain-instrinct-size: auto 90vw 380px`.
CleanShot 2022-03-04 at 10 35 49@2x

filipewl (commented):
Really cool CSS feature I didn't know much about. I learned a lot reading the web.dev link and other resources ([W3C](https://drafts.csswg.org/css-sizing-4/#intrinsic-size-override))! Interesting to see CSS Containment has been here since [Chrome 52](https://developers.google.com/web/updates/2016/06/css-containment), but only now seems to be getting traction. Maybe because there were some [accessibility concerns](https://web.dev/content-visibility/#a-note-on-accessibility) that only got addressed in 85-89 and improvements in [98](https://web.dev/content-visibility/#:~:text=of%20rendered%20content.-,In%20Chromium%2098,-and%20onward%2C%20there).

Left some minor suggestions and will re-review after reading your thoughts on them!

filipewl:
By the way, merging `master` should fix the CI issues addressed in https://github.com/vtex-sites/base.store/pull/369!

src/components/sections/Section/Section.scss
```diff
@@ -0,0 +1,9 @@
+.pages__section {
```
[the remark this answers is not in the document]
tlgimenes: Not sure. This is because the sections should be self contained and not use any global CSS. The way sections are being used today is only a wrapper to `src/components/ui` components, which is not ideal. I'll have another PR fixing this issue and making sections actually self-contained sections. 

The goal of sections is to integrate nicely with the CMS. The idea is that a section can be removed/added and their assets are also removed added from/to the page. That's why sections should be self contained.

src/components/sections/Section/Section.scss
tlgimenes: I thought about using `auto` too, but after watching this [video](https://www.youtube.com/watch?v=FFA-v-CIxJQ) from the Chrome team I realized that the first number does't really matter. In my tests, using `2px` for the width faster than auto, that's why I've left it in there like this.

src/components/sections/Section/Section.scss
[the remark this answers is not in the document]
filipewl: Ooh, I see! That's fine by me then.

src/components/sections/Section/Section.scss
filipewl: I see. What about the 1024px height, isn't that too much?

Out of curiosity, I was comparing the changes in this Layers tab on DevTools and noticed the height got too large, maybe because of that, not sure. Or that this height even matters much.
Before|After
-|-
Screen Shot 2022-03-03 at 19 11 15|Screen Shot 2022-03-03 at 19 10 30

I had bookmarked that video to watch later and learn more about containment. 
😆

filipewl: Just began watching the video and I understand now why the 1024px (or any other large value) for the height [doesn't seem to be a problem](https://youtu.be/FFA-v-CIxJQ?t=518)!

hellofanny: After watching the video and trying to figure out how it works, I'm coming here 🙈 Seems like it doesn't change much, and I haven't noticed any big issue related to the scroll, so I think it's okay to keep it like that.

eduardoformiga, line 4: I wonder if should we use the 
```suggestion
  contain-intrinsic-size: rem(2px) rem(1024px);
```

along with
```javascript
@import "src/styles/scaffold";
```

Just to keep the pattern if it didn't change the performance, though.

src/components/sections/Section/Section.tsx
```diff
@@ -0,0 +1,10 @@
+import './Section.scss'
```
eduardoformiga, line 1: Hmm, could we change to the singular just to keep the pattern?
```suggestion
import './section.scss'
```

hellofanny (approved):
Same thoughts as @filipewl! I didn't know much about it and took the chance to learn more. Thanks for the explanation! 🤩

src/pages/index.tsx
```diff
@@ -66,41 +67,49 @@ function Page(props: Props) {
...
-      </section>
+      </Section>
 
-      <section className="page__section page__section-shelf / grid-section">
-        <h2 className="title-section / grid-content">Most Wanted</h2>
-        <div className="page__section-content">
-          <ProductShelf products={products?.slice(0, 5)} />
-        </div>
-      </section>
+      <Section>
```
eduardoformiga, line 76: I'm not sure if the `Section` wrapper is the best approach since we are increasing the number of DOM nodes with `section`>`section`. Maybe we could transform the `Section.scss` into a sass `@mixin` and `@include` it elsewhere? Just thinking out loud, though.
OJl3rqu38P

eduardoformiga (commented):
Really nice to learn more about these CSS props! thanks for bringing it up. 🎉 

I left some comments just for the sake of understanding better the things

src/pages/index.tsx
hellofanny: Hmm. How about receiving the extra classes on the `<Section>` component? 

For example:

```
<Section  classes="grid-section grid-content">
    <h2 className="title-section">Just Arrived</h2>
    <div className="page__section-content">
        <ProductTiles products={products?.slice(5, 8)} />
    </div>
 </Section>
```

tlgimenes: Nice, now we have the same number of dom nodes :)

filipewl (approved):
🚀

PR #371, opened 2022-03-04
feat: sections

tlgimenes:
The whole idea of sections is to split app logic into replaceable/reusable strips of screen content so that CMS users can easily replace/rearrange the desired component on the screen. To allow CMS users to change the page's section, prismic and other CMSs usually use the following Render component logic:

```tsx
interface Section {
    name: string
    props: any
}

const Sections = {
   Hero: lazy(() => import('./sections/Hero'))
   BannerText: lazy(() => import('./sections/BannerText'))
   ProductShelf: lazy(() => import('./sections/ProductShelf'))
   // ... list goes on an on
}

function Render ({ sections }: { sections: Section[]}) {
  return (
    <>
      {sections.map(section => {
          const Component = Sections[section.name]
          
          return (
             <Suspense fallback={null}>
                <Component {...section.props} />
             <Suspense>
          )
      })}
    </>
  )     
}
```

This means that the `Section` map has to render a self contained section. 

This PR moves all external logic that were hard coded into the `/pages` components into the sections so CMS users can correctly change and re-order the sections

## How to test it?
Nothing should have changed

## Checklist
- [x] CHANGELOG entry added


igorbrasileiro:
Why this PR is not here @vtex-sites/storeframework.store ?

tlgimenes:
> Why this PR is not here @vtex-sites/storeframework.store ?

Because I want to have the less difference between this repo and storeframework.store. Without the need for CMS, the `sections` folder of this repo may be removed too

filipewl (approved):
LGTM! Found no visual regression on the main pages and the final organization on `pages/*` looks like what's expected (with only sections).

igorbrasileiro (approved):

src/components/sections/ProductGallery/ProductGalleryPage.tsx
```diff
@@ -78,7 +78,7 @@ function GalleryPage({
           />
           <div className="product-listing__results-sponsored">
             <h3>Sponsored</h3>
-            <ProductTiles products={productsSponsored.slice(0, 2)} />
+            <ProductTiles products={productsSponsored.slice(0, 2)} title="" />
```
eduardoformiga, line 81: Should we make the title prop optional?

eduardoformiga (approved):
Nice Job! 🎉 
should we remove the `Section` from the [search page](https://github.com/vtex-sites/base.store/blob/master/src/pages/s.tsx#L64-L66) as well?

tlgimenes:
> 

Nice! I haven't seen that! I'll change it

src/components/sections/ProductGallery/ProductGalleryPage.tsx
tlgimenes: Not sure. This component is weird because it's using another section. Sections should be self contained and not depend on other sections. This should be refactored on another PR. I'll add a TODO for this. Thanks!

PR #372, opened 2022-03-04
feat: Search Suggestions component

lucasfp13:

This PR intends to add the `Suggestions` component. The purpose of this component is to encapsulate autocomplete suggestions and suggested products.

Note: The component has mock data as the integration will be done later.

## How does it work?

Screen Shot 2022-03-10 at 11 24 32

## How to test it?

Since the component will be part of the `Suggestion`'s dropdown component, we don't have a specific place to render the component yet.

But one option is to render it in the `Navbar` component as shown below:

```
// line 84
<SearchInput
  onBlur={() => setSearchExpanded(false)}
  onFocus={() => setSearchExpanded(true)}
/>
```

```
// line 110
{searchExpanded && <Suggestions />}
```
Screen Shot 2022-03-10 at 11 32 37
Screen Shot 2022-03-10 at 11 32 50

## References

[FSSS 191 - Search Suggestions component](https://vtex-dev.atlassian.net/browse/FSSS-191)

## Checklist

- [x] CHANGELOG entry added

### DIFF
--- a/src/components/search/Suggestions/Suggestions.tsx
+++ b/src/components/search/Suggestions/Suggestions.tsx
@@ -57,6 +57,30 @@ function formatSearchTerm(
   return searchTerm.toLowerCase()
 }
 
+function handleSuggestions(suggestion: string, searchTerm: string) {
+  const suggestionSubstring = suggestion
+    .toLowerCase()
+    .split(searchTerm.toLowerCase())
+
+  return (
+    <p>
+      {suggestionSubstring.map((substring, indexSubstring) => (
+        <>
+          {substring.length > 0 && (
+            <b className="suggestions__item-bold">
+              {indexSubstring === 0
+                ? substring.charAt(0).toUpperCase() + substring.slice(1)
+                : substring}
+            </b>
+          )}
+          {indexSubstring !== suggestionSubstring.length - 1 &&
+            formatSearchTerm(indexSubstring, searchTerm, suggestion)}
+        </>
+      ))}
+    </p>
+  )
+}
+
 export interface SuggestionsProps extends HTMLAttributes<HTMLDivElement> {
   /**
    * ID to find this component in testing tools (e.g.: cypress, testing library, and jest).
@@ -77,30 +101,6 @@ const Suggestions = forwardRef<HTMLDivElement, SuggestionsProps>(
       SUGGESTED_PRODUCTS.length > 0
         ? SUGGESTIONS.slice(0, MAX_SUGGESTIONS_WITH_PRODUCTS)
         : SUGGESTIONS.slice(0, MAX_SUGGESTIONS)
-
-    const handleSuggestions = (suggestion: string, searchTerm: string) => {
-      const suggestionSubstring = suggestion
-        .toLowerCase()
-        .split(searchTerm.toLowerCase())
-
-      return (
-        <p>
-          {suggestionSubstring.map((substring, indexSubstring) => (
-            <>
-              {substring.length > 0 && (
-                <b className="suggestions__item-bold">
-                  {indexSubstring === 0
-                    ? substring.charAt(0).toUpperCase() + substring.slice(1)
-                    : substring}
-                </b>
-              )}
-              {indexSubstring !== suggestionSubstring.length - 1 &&
-                formatSearchTerm(indexSubstring, searchTerm, suggestion)}
-            </>
-          ))}
-        </p>
-      )
-    }
 
     return (
       <section
--- a/src/components/search/Suggestions/Suggestions.tsx
+++ b/src/components/search/Suggestions/Suggestions.tsx
@@ -112,11 +112,8 @@ const Suggestions = forwardRef<HTMLDivElement, SuggestionsProps>(
         {suggestions.length > 0 && (
           <UIList data-suggestions-list className="suggestions__section">
             {suggestions?.map((suggestion, index) => (
-              <li key={index}>
-                <Button
-                  onClick={() => null}
-                  className="suggestions__suggestion"
-                >
+              <li key={index} className="suggestions__item">
+                <Button onClick={() => null}>
                   {handleSuggestions(suggestion, term)}
                 </Button>
               </li>
@@ -130,7 +127,7 @@ const Suggestions = forwardRef<HTMLDivElement, SuggestionsProps>(
             <UIList>
               {SUGGESTED_PRODUCTS.slice(0, MAX_SUGGESTED_PRODUCTS).map(
                 (product, index) => (
-                  <li key={index}>
+                  <li key={index} className="suggestions__item">
                     <SuggestionProductCard product={product} />
                   </li>
                 )
--- a/src/components/search/Suggestions/Suggestions.tsx
+++ b/src/components/search/Suggestions/Suggestions.tsx
@@ -1,0 +1,108 @@
+import type { HTMLAttributes } from 'react'
+import React, { forwardRef } from 'react'
+import { List as UIList } from '@faststore/ui'
+import Link from 'src/components/ui/Link'
+
+import SuggestionProductCard from '../SuggestionProductCard'
+
+import './suggestions.scss'
+
+const MAX_SUGGESTED_PRODUCTS = 4
+const SUGGESTED_PRODUCTS = [
+  {
+    name: 'Ergonomic Wooden Bacon',
+    listPrice: 72.06,
+    price: 46.26,
+    image: [
+      {
+        alternateName: 'rerum',
+        url: 'http://storeframework.vtexassets.com/arquivos/ids/167285/ut.jpg?v=637753017045600000',
+      },
+    ],
+  },
+  {
+    name: 'Handcrafted Rubber Sausages',
+    listPrice: 59.57,
+    price: 32.83,
+    image: [
+      {
+        alternateName: 'ea',
+        url: 'http://storeframework.vtexassets.com/arquivos/ids/155949/voluptas.jpg?v=637752878341070000',
+      },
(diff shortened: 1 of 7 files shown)


src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    return (
+      <section
+        ref={ref}
+        data-testid={testId}
+        data-store-suggestions
+        className="suggestions"
+        {...otherProps}
+      >
+        <UIList data-suggestions-list className="suggestions__section">
```
filipewl, line 106: We need to wrap the children in `li`s because list elements such as these `ul`s should only have `li` as children, nothing else.

CleanShot 2022-03-10 at 14 14 55@2x

It's the same for the list with suggested products in the lines below.

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+          {suggestions.map((suggestion, index) => (
+            <Button
```
filipewl, line 108: Something for us to think about when we have more details about the implementation in the future, but I could see this also being a link.

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const Suggestions = forwardRef<HTMLDivElement, SuggestionsProps>(
+  function Suggestions(
+    { testId = 'suggestions', term = '', ...otherProps },
+    ref
+  ) {
+    const suggestions =
+      SUGGESTED_PRODUCTS.length > 0
+        ? SUGGESTIONS.slice(0, MAX_SUGGESTIONS)
+        : SUGGESTIONS
```
filipewl, line 58: Isn't there any limit to the number of suggestions when there are no suggested products?

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+                      : char.toLowerCase()
+                  )
+              }
+
+              return searchTerm.toLowerCase()
+            }
+
+            return (
+              <>
+                <b className="suggestions__suggestion-bold">
```
filipewl, line 84: It doesn't affect anything but noticed some times this element is rendered without children:
CleanShot 2022-03-10 at 14 43 50@2x

filipewl (commented):
Looking good to me so far! I'll have to get back to it later.

src/components/search/Suggestions/Suggestions.tsx
eduardoformiga: Nice catch! according to Figma, the limit when there are no suggested products is 10 suggestions.

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+          {suggestions.map((suggestion, index) => (
+            <Button
+              key={index}
+              onClick={() => null}
+              className="suggestions__suggestion"
+            >
+              {handleSuggestions(suggestion, term)}
+            </Button>
+          ))}
+        </UIList>
```
eduardoformiga, line 116: Should we check the `suggestions` existence before rendering?
```suggestion
        {suggestions.length > 0 && (
          <UIList data-suggestions-list className="suggestions__section">
            {suggestions.map((suggestion, index) => (
              <Button
                key={index}
                onClick={() => null}
                className="suggestions__suggestion"
              >
                {handleSuggestions(suggestion, term)}
              </Button>
            ))}
          </UIList>
        )}
```

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+        <div className="suggestions__section">
+          <p className="suggestions__title">Suggested Products</p>
+          <UIList>
+            {SUGGESTED_PRODUCTS.slice(0, MAX_SUGGESTED_PRODUCTS).map(
+              (product, index) => (
+                <SuggestionProductCard key={index} product={product} />
+              )
+            )}
+          </UIList>
+        </div>
```
eduardoformiga, line 127: the same here for the `SUGGESTED_PRODUCTS` (when we have the integration done will be easier to deal with it).
```suggestion
        {SUGGESTED_PRODUCTS.length > 0 && (
          <div className="suggestions__section">
            <p className="suggestions__title">Suggested Products</p>
            <UIList>
              {SUGGESTED_PRODUCTS.slice(0, MAX_SUGGESTED_PRODUCTS).map(
                (product, index) => (
                  <SuggestionProductCard key={index} product={product} />
                )
              )}
            </UIList>
          </div>
        )}
```

eduardoformiga (commented):
Agreed with @filipewl suggestions.
I will review the `handleSuggestions` deeply afterward, but so far left some questions.

src/components/search/Suggestions/Suggestions.tsx
eduardoformiga: I think that should be a link if we have page navigation involved.

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+              return searchTerm.toLowerCase()
+            }
+
+            return (
+              <>
+                <b className="suggestions__suggestion-bold">
+                  {indexSubstring === 0
+                    ? substring.charAt(0).toUpperCase() + substring.slice(1)
+                    : substring}
+                </b>
```
eduardoformiga, line 88: We could check the substring existence before rendering for the sake of avoiding the element that @filipewl mentioned above.
```suggestion
                {substring.length > 0 && (
                  <b className="suggestions__suggestion-bold">
                    {indexSubstring === 0
                      ? substring.charAt(0).toUpperCase() + substring.slice(1)
                      : substring}
                  </b>
                )}
```

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+                  .split('')
+                  .map((char, idx) =>
+                    idx === 0 && suggestion.indexOf(char.toUpperCase()) === 0
+                      ? char.toUpperCase()
+                      : char.toLowerCase()
+                  )
+              }
+
+              return searchTerm.toLowerCase()
+            }
```
eduardoformiga, line 80: What do you think about moving this function outside the Component block, so that we can avoid creating the function on every `map` iteration, then we can inject the dependencies? like:
```javascript
formatSearchTerm(indexSubstring, searchTerm, suggestion)
```

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+          {suggestionSubstring.map((substring, indexSubstring) => {
+            const formatSearchTerm = () => {
+              if (indexSubstring === 0) {
+                return searchTerm
+                  .split('')
+                  .map((char, idx) =>
+                    idx === 0 && suggestion.indexOf(char.toUpperCase()) === 0
+                      ? char.toUpperCase()
+                      : char.toLowerCase()
+                  )
```
eduardoformiga, line 76: We could `join` the [splited chars](https://github.com/vtex-sites/base.store/pull/372/files#r823990440):
```suggestion
                  )
                  .join('')
```

src/components/search/Suggestions/Suggestions.tsx
lucasfp13: I have discussed it with @hellofanny too, we agreed that we need to validate this when integrating this component with a closer look at what we get from the server.

src/components/search/Suggestions/suggestions.scss
```diff
@@ -29,3 +44,18 @@
...
+
+[data-store-suggestions] {
+  .suggestions__suggestion-bold {
+    color: var(--color-neutral-7);
+    font-weight: var(--text-weight-bold);
+  }
+
+  .button[data-store-button] {
+    padding: var(--space-2) 0;
+    color: var(--color-text-subtle);
```
hellofanny, line 56:
```suggestion
    color: var(--color-text-subtle);
    font-weight: var(--text-weight-regular);
```

src/components/search/Suggestions/suggestions.scss
```diff
@@ -29,3 +44,18 @@
     }
   }
 }
+
+[data-store-suggestions] {
+  .suggestions__suggestion-bold {
```
hellofanny, line 49: How about called it "suggestions__item-bold"?

hellofanny (approved):
Well done! 🚀  I've left just one more suggestion 👍

src/components/search/Suggestions/Suggestions.tsx
```diff
@@ -0,0 +1,146 @@
...
+  function Suggestions(
+    { testId = 'suggestions', term = '', ...otherProps },
+    ref
+  ) {
+    const suggestions =
+      SUGGESTED_PRODUCTS.length > 0
+        ? SUGGESTIONS.slice(0, MAX_SUGGESTIONS_WITH_PRODUCTS)
+        : SUGGESTIONS.slice(0, MAX_SUGGESTIONS)
+
+    const handleSuggestions = (suggestion: string, searchTerm: string) => {
```
filipewl, line 81: Could this also be considered a functional component (because it is returning some elements) instead of just a function?

Do you know if there's any penalty in defining it inside another component or if it's best to define it outside?

filipewl (approved):
LGTM, but I'm in doubt about `handleSuggestions`, see my comment!

eduardoformiga (approved):

sergiofontes (approved):

PR #373, opened 2022-03-07
[Gatsby v4] Add support for TypeScript files to configure GraphQL servers (/src/server folder) and migrate config files to TypeScript

victorhmp:

This PR aims to increase the usage of TypeScript in the codebase, and the main feature added was the transition from using `.js` files in the `/src/server` folder to using `.ts` files. This should enable users to extend the default server configuration and GraphQL Schema using TypeScript files, which are not supported out of the box today.

This PR also allows us to use TypeScript in the following Gatsby specific files:

- `gatsby-config`
- `gatsby-ssr`
- `gatsby-browser`
- `gatsby-node`

Some minor changes also implemented:

- Added a missing developer dependency: `@lhci/cli`. This should be present in the `package.json` file since our `lhci` script depends on it.
- Fixed type inconsistencies between types defined in `@faststore/sdk` and how they were being used here.

## How to test it?

Since this is more of a structural change, to make sure everything is working, all PR checks should be working and the deploy preview should be no different from what's currently live.

## Checklist

- [] CHANGELOG entry added


victorhmp:
The deploy-preview for this won't be up because the base branch here is not `master`.

Also, the build doesn't pass on WebOps, but it does build correctly, just like the `chore/gatsby-v4` branch does:

Screen Shot 2022-03-07 at 14 07 49

igorbrasileiro (approved):

gatsby-ssr.tsx
```diff
@@ -11,7 +11,17 @@ import TestProvider from './src/sdk/tests'
...
 
-export const wrapRootElement = ({ element }) => (
+// Gatsby types the returned elements from `getHeadComponents` as
+// `React.ReactNode`, but this is inaccurate. The attributes defined below
+// are present in those elements.
+interface HeadComponents
+  extends ReturnType<PreRenderHTMLArgs['getHeadComponents']> {
+  type: string
+  key: string
+  props?: Record<string, unknown>
```
victorhmp, line 21: @igorbrasileiro answering to the comment you left on the previous PR, the other types that could be used as keys in the `props` object are already accounted for here. `Record` is defined as: `type Record<K extends string | number | symbol, T> = { [P in K]: T; }`

PR #375, opened 2022-03-07
chore: Improve images

tlgimenes:
Improves image processing by using great thumbor features. 

## How does it work?
Thumbor has amazing tools regarding image cropping and resizing. To know more: https://thumbor.readthedocs.io/en/latest/usage.html#image-endpoint.  
We were extensively using one of these features called `fit-in`. This property is ok when the images behave badly, however, our UI thinks the images are nice and have the same aspect ratio. In this case we can drop out of using `fit-in` and save some bytes out of the images.

Also, another optimization this PR brings is to render img components once, by moving the creation of the `breakpoints` array outside the rendering cycle of the app. This should save some re-renders too.

Finally, hero's aspect-ratio image was wrong. On moto g4 the right aspect ratio was 3/2, not 2. The fix makes the browser work less to fit the image that comes from the server with the image that will be rendered on the screen. Below, a comparison of the images coming from the thumbor server. The second one makes the potato (moto g4) work much less than one the first one, because it has the same aspect-ratio of the final rendered image.

<div style="display: flex">
image
image
</div>

## How to test it?
Everything should work as before

## Checklist
- [x] CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `EmptyState` at the `ProductGallery` section.
 
 ### Changed
-
+- Removed fit-in property from image component
 - Moves icons to `/static/icons` folder
 - Replaces page type redirects, a.k.a. `/account`, `/login` to a corresponding file in `/pages` folder
 - Replaces `let` declarations for `useRef` for better React compatibility
--- a/src/components/cart/CartItem/CartItem.tsx
+++ b/src/components/cart/CartItem/CartItem.tsx
@@ -16,6 +16,15 @@ interface Props {
   item: ICartItem
 }
 
+const imgOptions = {
+  sourceWidth: 360,
+  aspectRatio: 1,
+  width: 72,
+  breakpoints: [50, 100, 150],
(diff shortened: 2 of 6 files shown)


filipewl (commented):
Since [the 1st commit](https://github.com/vtex-sites/base.store/pull/375/commits/969fd022fcf04793a7162a516316872d6bb11b8c), [LHCI](https://github.com/vtex-sites/base.store/runs/5455575862) complained about not properly sizing the Hero image for the PLP and I can't understand why. Other than [the `fitIn` removal](https://github.com/vtex-sites/base.store/pull/375/commits/969fd022fcf04793a7162a516316872d6bb11b8c#diff-73517a53d044dea02bd41ebb6348c2e123068b83e26b834308dbe43e736d0cb1L70), I could only spot the addition of [the breakpoints](https://github.com/vtex-sites/base.store/pull/375/commits/969fd022fcf04793a7162a516316872d6bb11b8c#diff-73517a53d044dea02bd41ebb6348c2e123068b83e26b834308dbe43e736d0cb1R37).

These changes look good to me, but do you have any leads in the LHCI problem?

filipewl (approved):
LGTM now!

PR #376, opened 2022-03-08
fix: image padding

tlgimenes:
Quick fix removing unecessary padding


igorbrasileiro:
Still have problems with padding ( 3 / 2 = means 2/3 = 66.6666...%, so the aspect ratio value is right, but the padding isn't), i think it's related to the wrong aspect ratio. Have you ever tried to use two different images, combining them with sizes and srcset?

I suggest that because the image 3/2 fix the mobile aspect ratio, but this aspect ratio isn't right for the desktop.

igorbrasileiro (approved):

igorbrasileiro:
We decided to use this padding without right value on desktop, that has more power computation

PR #377, opened 2022-03-08
chore: Remove hooks folder

tlgimenes:
This PR removes the hooks folder. 

## How does it work?
The main idea of the `sdk` folder is to contain hooks with business logic that are used across the app. Turns out a new folder `hooks` was created, which lead to duplication of purposes on the filesystem. 
This PR unifies both hooks and sdk folders. Initially, I thought about moving the files "as is" from hooks to sdk, however, a closer inspection revealed that hooks on the `hooks` folder were only being used in a single file. I then, decided to inline these functions with the intent of having a smaller code base.

## How to test it?
Everything should work as before

## Checklist
- [x] CHANGELOG entry added


src/components/common/PostalCode/PostalCodeInput.tsx
```diff
@@ -1,21 +1,22 @@
...
+const POSTAL_CODE_STORAGE_KEY = 'main::store::postalCode'
 const POSTAL_CODE_INPUT_ID = 'postal-code-input'
 
 export default function PostalCodeInput() {
-  const [localPostalCode, setLocalPostalCode] = useState<string>('')
-  const [, setPostalCode] = usePostalCode()
-  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
-    setLocalPostalCode(event.target.value)
-  }
+  const [code, setPostalCode] = useStorage<string>(POSTAL_CODE_STORAGE_KEY, '')
```
filipewl, line 12: Just to better indicate that these variables are related, what do you think of renaming `code` to `postalCode` or `setPostalCode` to `setCode`?

filipewl (approved):
LGTM! I remember a previous comment where you mentioned this need to move everything from `/hook` to `/sdk`.

eduardoformiga (approved):
LGTM! Moving the `useSearchParams.ts` inline looks harder for readability. But since we just use it on one file and it's good to improve the codebase size, I agree with this change :)

CHANGELOG.md
```diff
@@ -36,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - useWindowDimensions hook
 
 ### Removed
-
+- Hooks folder
```
igorbrasileiro, line 39: I think we could say "removing hooks folder and migrating these hooks to sdk ou inline them on components"

src/components/common/PostalCode/PostalCodeInput.tsx
igorbrasileiro: This hook was planned to be used in other components/hooks. But we was doing the feature step by step

src/components/common/PostalCode/PostalCodeInput.tsx
```diff
@@ -1,21 +1,22 @@
...
+  const ref = useRef<HTMLInputElement>(null)
 
   const handleSubmit = (event: KeyboardEvent<HTMLInputElement>) => {
-    if (event.key === 'Enter' && Boolean(localPostalCode)) {
-      setPostalCode(localPostalCode)
+    const val = ref.current?.value
```
victorhmp, line 16:
```suggestion
    const value = ref.current?.value
```

or 

```
const currentValue = ref.current?.value
```

victorhmp (changes requested):

victorhmp (approved):

PR #379, opened 2022-03-09
[FSSS-200] fix: Remove product card border on mobile

heitorado:
This PR removes the product card's border when rendering the component on mobile devices.

## How does it work?
Since the order that CSS loads is non-deterministic, we have to make sure that even if we shuffle the stylesheets, the styles are still being applied correctly. For achieving the final result, the media query was removed from the `product-grid.scss` file and included in both `product-card.scss` and `product-card-skeleton.scss`. 
By doing that we ensure that the styles aplied by the `product-card` stylesheets are not overwritten by the `product-grid` stylesheet when the loading order changes.
Following the mobile-first approach, we make the product cards borderless by default, and then add the `@include media(">=notebook")` media query into the product card's stylesheets to add borders on larger screen sizes. 

## How to test it?

- Open a product listing page on a larger screen, check that the product cards have borders.
- Switch to responsive layout in a smaller screen and check that the product cards are now borderless.

## References

Thanks to @hellofanny and @filipewl for the help.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+
 - Sections component with `content-visibility: auto`
 - Webpack Bundle analyzer
 - `GatsbyLink` to `Link` ui component.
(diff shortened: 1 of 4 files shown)


filipewl (approved):
Yay! I think this should fix the problem from recurring.

hellofanny (approved):
Yay! I've tested here, and it seems to fix this issue. 🙃 Although we are repeating the same code in two places, I think it's clear to leave it like that. Maybe later, we can consider another approach for the card and its skeleton.

eduardoformiga (approved):
Good Job! 🎉 

I also think that the style fits better placed on the `product-card` and its `skeleton` instead of the `product-grid`. 

Preview ✨ 
nGJLWOO6fV

PR #380, opened 2022-03-10
[FSSS-180] Fix: facets not being selected

eduardoformiga:

While using the mobile version of the filter on PLP, sometimes the facets were not being applied as we can see below:

https://user-images.githubusercontent.com/11325562/157688437-d4ac1b4c-718b-49fe-b31d-c41ae0b811f6.mp4

So, this PR aims to look into this problem and fix it.

## How does it work?

The `Filter` on the mobile version works in a different way than the desktop version.
While on the desktop the result is shown just after the user clicks the `Facet's Checkboxes`, performing the `toggleFacet` (singular, so one facet by time) function from the `useSearch` hook (`@faststore/sdk`). 

On the mobile version, the `Filter` opens inside a `SlideOver` component and the user first picks up the facets (`onFacetChange` function) and so hits the _Apply_ `Button` (`onApply` function). Thus, we need to investigate these 2 functions.

The `onFacetChange` aims to keep the actual state once the user selects/unselects the checkboxes. 

Previously 2 states were being used: `selectedFacets` and `facetsToRemove`. But keeping the `facetsToRemove` can be a little tricky. Once the facet is already selected and the user unselects it, the facet was being added to the `facetsToRemove` array. But the opposite situation, when the facet was unselected and the user selects it again, the  `facetsToRemove` state wasn't updated. 

If the user selects and unselects the same facet multiple times without hitting the _Apply_ `Button`, the final `facetsToRemove` array is gonna have the same facet multiples times. Making the end result messy afterward on the `onApply` function. 

Furthermore, we don't need to keep the `facetsToRemove` state, once we have the original state `searchState.selectedFacets` and the actual state `selectedFacets`, we can get which facets were added and removed comparing these 2 arrays against each other.

So, [the first task](https://github.com/vtex-sites/base.store/pull/380/commits/d2f03a2aa843ba5431b345a7ea7fcd469d11edeb) was remove the `facetsToRemove` state.

[I took advantage](https://github.com/vtex-sites/base.store/pull/380/commits/ecb5a99021474e90f8a8309a996058725b953ccd) that I was going to change this function and refactor it a little bit trying to improve the readability and performance. Besides, I changed the `setSelectedFacets` to the [functional update](https://reactjs.org/docs/hooks-reference.html#functional-updates) way, recommended by React doc when we depend on previous values.

Another problem is the referential equality while using the `searchState.selectedFacets` from the `@faststore/sdk` to [set the initial value](https://github.com/vtex-sites/base.store/pull/380/commits/60e843af7c201ffd27b6a9021dfd8c5d87749703) of the `selectedFacets` and also every time [we open the filter](https://github.com/vtex-sites/base.store/pull/380/commits/ab78620df16ae36d260c8548540839c8a9884be7). The first time we `setSelectedFacets` on  `onFacetChange` function was also changing the `searchState.selectedFacets`. To fix it we can use a new array spreading the `searchState.selectedFacets` values.

Finally, on the `onApply` function, we need to use the `toggleFacets` (plural, multiple facets) function from the `@faststore/sdk` to toggle all facets changed (added and removed). Now that we clean the tricky `facetsToRemove`, that was duplicating facets and making the messy. We need to compare the original state `searchState.selectedFacets` and the actual state `selectedFacets` as mentioned before to get the Facets added and removed (like an XOR operation) and passing the results to the `toggleFacets` ([task done here](https://github.com/vtex-sites/base.store/pull/380/commits/7194dcb379bd4003e6a81f4553caa6d25ac3ad05)).

Result:

https://user-images.githubusercontent.com/11325562/157735910-08452ccf-0f86-4b4c-b99a-2f42ad5a549c.mov

## How to test it?

You can open the [preview](https://preview-380--base.preview.vtex.app/office/) and double-check if the filter works as expected while using the mobile version.

## Checklist

- [x] CHANGELOG entry added


src/components/search/Filter/Filter.tsx
```diff
@@ -145,13 +140,26 @@ function Filter({
...
+    })
+
+    const facetsToRemove = searchState.selectedFacets.filter((facet) => {
+      const isSameFacet = (f: IStoreSelectedFacet) =>
+        f.value === facet.value && f.key === facet.key
+
+      const isFacetSelected = selectedFacets.some(isSameFacet)
+
+      return !isFacetSelected
+    })
```
filipewl, line 160: What do you think of extracting a function here? It's the same logic, but for different instances of selected facets.

filipewl (approved):
Awesome PR description! It guided me commit-by-commit through the changes. 
🤩

I tested on my Android and also confirmed the bug is now gone.

src/components/search/Filter/Filter.tsx
eduardoformiga: That's true! I'm going to implement it. Thanks!

lucasfp13 (approved):
Excellent context description! 💯 

I tested it again to force the error but everything seems fine now!

filipewl (approved):

igorbrasileiro (approved):

PR #383, opened 2022-03-11
refactor: `Tiles` component to use a semantic list element

filipewl:

The `Tiles` component renders a list of `Tile` components, so it's essentially a list, but it's using generic `div's. 

## How does it work?

We can improve its semantic markup using `ul` &`li`.

## How to test it?

This component isn't rendered on any page, so there isn't an easy way to test it, but the changes are pretty straightforward.

Example of how these components are used:
https://github.com/vtex-sites/base.store/blob/0924deb195d9dbb0c0220378c79ee057d397c85b/src/components/sections/ProductTiles/ProductTiles.tsx#L30-L34

CleanShot 2022-03-11 at 15 30 20

## References

- [FSSS-203 Refactors Tiles component to use List](https://vtex-dev.atlassian.net/browse/FSSS-203)


eduardoformiga (approved):
Good Job! 👏

hellofanny (approved):
Niceeeeee ❤️

src/components/ui/Tiles/Tile.tsx
```diff
@@ -1,22 +1,22 @@
 import React, { forwardRef } from 'react'
 import type { HTMLAttributes } from 'react'
 
-export interface TileProps extends HTMLAttributes<HTMLDivElement> {
+export interface TileProps extends HTMLAttributes<HTMLLIElement> {
```
hellofanny, line 4: At first I thought it was a “typo" (LI uppercase) hahaha, but it's really like [that](https://developer.mozilla.org/en-US/docs/Web/API/HTMLLIElement). 🙈

PR #386, opened 2022-03-14
chore: Homogeneous codebase

tlgimenes:
Makes codebase homogeneous by:
1. Moving IconSVG to `ui` folder.
2. Moving footer svg icons from custom files to `icons.svg` file
3. Using `IconSVG` in footer component
4. Moving EmptyState component to `ui` folder.

## How does it work?
This PR revolves around having one example on the codebase for both "where placing UI components" and "how to add icons to the store". We should have only one solution for both tasks. Currently, we have two ways of doing these tasks. This PR solves this problem.

As a side effect, this PR removed some code from the repo: 
image

## How to test it?
Everything should be the same

## Checklist
- [x] CHANGELOG entry added


src/components/common/Footer/footer.scss
```diff
@@ -93,6 +93,11 @@
       }
     }
 
+    svg {
+      border: 1px solid #bbc1c6;
+      border-radius: 3px;
```
filipewl, line 98: Some tokens I think are similar:

```suggestion
      border: var(--border-width-0) solid var(--color-neutral-3);
      border-radius: var(--border-radius-small);
```

filipewl (approved):
Loved the shorter naming and that more icons were added.

I ask only to also take a look at [Managing SVG Icons](https://github.com/vtex-sites/base.store/tree/fb79cd6aa99f9fa0fa6003dc6f2117abb30f014b#managing-svg-icons) section on the README to update some references (e.g. component name, static file location).

igorbrasileiro (approved):

igorbrasileiro:
@tlgimenes i know this isn't related to this PR, but, can you add 18px width to SignInLink icon? This is causing a Layout Shift

tlgimenes:
> @tlgimenes i know this isn't related to this PR, but, can you add 18px width to SignInLink icon? This is causing a Layout Shift

AFAIK, it's already with 18px

PR #389, opened 2022-03-15
feat: Add VTEX Search tracking

tlgimenes:
This PR integrates with VTEX Search tracking API. Also, this PR integrates two events:
1. search.query
2. search.click

These events are necessary for populating `Top searches` and adding relevance to search. 

Also, this PR adds the origin url to all analytics events so we can know which url generated each event.

## How does it work?
Search does not have a proper script, like VTEX RC or Google Analytics, but a tracking API. With that, I have implemented a client for this tracking api on the file `src/sdk/analytics/platform/vtex/search.ts`. I have roughly followed [the docs](https://www.notion.so/vtexhandbook/Event-API-Documentation-48eee26730cf4d7f80f8fd7262231f84) on which fields we send to the tracking api. 

After doing the client, I had to raise the right events and capture them to send to the tracking API. Turns out, we already had both events, 'search' and 'select_item' being raised.

`search` event was being raised inside the `doSearch` function. We don't have the necessary info inside this function, so I've moved this event to the `ProductGallery`. This change, made `search` event to be raised on the following situations:
1. When the user search a term via the searchbar.
2. When (inside the search result page), the user changes the sorting method.
3. When (inside the search result page), the user changes the filter applied.

I'm not sure if we should raise the event on cases 2,3. Make comments on the review!

Finally, I thought it would be interesting to add a url for each raised event, so our analytics engine can enrich event data with the current url the user is viewing. I see this being useful to track the urls where the user adds/remove a product to/from cart etc.

Also, make sure to review:
https://github.com/vtex/faststore/pull/1174

## How to test it?
To track search events, open the network tab of chrome, filter by `sp.vtex.com`. Them search by terms and click on products. You should see requests being made.
Make sure these events are only raised on full text pages, not on department pages or shelves.

## References
- Search tracking API Docs: https://www.notion.so/vtexhandbook/Event-API-Documentation-48eee26730cf4d7f80f8fd7262231f84

## Checklist
- [x] CHANGELOG entry added


src/sdk/analytics/platform/vtex/search.ts
```diff
@@ -0,0 +1,70 @@
...
+  | {
+      text: string
+      match: number
+      misspelled?: boolean
+      operator: 'and' | 'or'
+      url: string
+      type: 'search.query'
+    }
+
+export const send = (options: SearchEvent & { url?: string }) =>
```
igorbrasileiro, line 58: How about add this function to the party town?

src/components/sections/ProductGallery/ProductGallery.tsx
```diff
@@ -24,14 +26,33 @@ interface Props {
...
   const totalCount = data?.search.products.pageInfo.totalCount ?? 0
   const { next, prev } = usePagination(totalCount)
 
-  if (data && totalCount === 0) {
+  /**
+   * Send analytics event every time a search occurs.
+   * Search engines use this event for populating `topSearches` indices
+   * */
+  useEffect(() => {
+    if (searchState.term != null && !isLoading) {
```
igorbrasileiro, line 43: Use strict comparator

src/sdk/analytics/platform/vtex/search.ts
tlgimenes: How can we do it?

igorbrasileiro: The party town proxy functions inside the window. So can transform it as a global function, or transform it in a script inside JS file and fetch it through  tag `script` .

tlgimenes: If we do this, we will need a script tag and move this `send` function into this script, right? I'm not sure if we need to do this since this function is hard-coded by us and pretty small and lazy-loaded

src/sdk/analytics/platform/vtex/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  window?.sendrc?.(event.name, event.params)
+
+  // VTEX Intelligent Search
+  if (event.name === 'search') {
+    send({
+      type: 'search.query',
+      text: event.params.search_term,
+      match: event.params.records_filtered,
+      url: event.params.url,
+      operator: 'and',
```
chrsmutti, line 24: `operator` is metadata that gets sent back on each `product_search` response. Is it possible to grab the returning value and set it correctly here?

Example below:
https://search.biggylabs.com.br/search-api/v1/storecomponents/api/split/product_search/?q=black%20top%20shirt&c=1

```
  "operator": "or",
```

There's also an unmeasured (we can run some tests later) performance increase when you keep metadata information between page changes and facets selection, such as `operator` and `fuzzy`, but we can talk about it on another occasion!

src/sdk/analytics/platform/vtex/index.ts
```diff
@@ -0,0 +1,48 @@
...
+      const productId = item.item_id ?? item.item_variant
+      const position = item.index
+
+      if (productId && position) {
+        send({
+          type: 'search.click',
+          productId,
+          position,
+          url: url.href,
+          text: url.searchParams.get('q') ?? 'uknown',
```
chrsmutti, line 43: If this can happen, can we change it to `<empty>`? This used to be the data that got sent by our SDK on such cases, so there are already systems in place to handle that.

```suggestion
          text: url.searchParams.get('q') ?? '<empty>',
```

chrsmutti (changes requested):
Other than some undocumented quirks of the event-api (:my-bad:), the code looks good to me, but please take a look on the comments I left. 💪

src/sdk/analytics/platform/vtex/search.ts
igorbrasileiro: Ok, no problem, my main concern is about the functions to create random UUID and other small stuff. But that's ok

src/sdk/analytics/platform/vtex/index.ts
tlgimenes: Yeah, we don't really have access to these values in here. Also, since we don't add these as parameters to the search, I think it's ok to set them as the default used on search. If you prefer, we can remove these operator and fuzzy all together.

src/sdk/analytics/platform/vtex/search.ts
```diff
@@ -0,0 +1,70 @@
+/**
+ * More info at: https://www.notion.so/vtexhandbook/Event-API-Documentation-48eee26730cf4d7f80f8fd7262231f84
```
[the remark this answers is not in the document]
tlgimenes: you mean move all content from notion to here? Search API should have a public doc. Once they release something we can update in here, what do you think?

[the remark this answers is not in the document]
igorbrasileiro: Yeah, sounds good to me!

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-
+- Adds vtex search tracking script. With this we will populate TopSearches and Autocomplete indices
```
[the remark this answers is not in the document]
igorbrasileiro: Are you adding the word `add` twice, since this section is `Added`

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add hideUnavailableItems at store.config
+- Adds vtex search tracking script. With this we will populate TopSearches and Autocomplete indices
```
igorbrasileiro, line 13:
```suggestion
```

src/sdk/analytics/platform/vtex/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        send({
+          type: 'search.click',
+          productId,
+          position,
+          url: url.href,
+          text: url.searchParams.get('q') ?? '<empty>',
+        })
+      }
+    }
+  }
```
igorbrasileiro, line 46: To make the sendEvent function a little bit clearer, can you move this code to a new function?

src/sdk/analytics/platform/vtex/search.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import config from 'src/../store.config'
+
+const THIRTY_MINUTES_S = 30 * 60
+const ONE_YEAR_S = 365 * 24 * 3600
```
igorbrasileiro, line 8:
```suggestion
const ONE_YEAR_S = 365 * 24 * 3_600
```

igorbrasileiro: Or you can just multiply 60 secs * 60 min

src/sdk/analytics/platform/vtex/search.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const randomUUID = () =>
+  typeof crypto.randomUUID === 'function'
+    ? crypto.randomUUID()
+    : (Math.random() * 1e6).toFixed(0)
+
+const storage = (key: string, expiresSecond: number) => {
```
igorbrasileiro, line 15: Rename this storage name to a better one, this function is creating a new storage handler for a specific key, wdyt  "storageFactory" or "storageCreator"?

src/sdk/product/useProductLink.ts
```diff
@@ -1,5 +1,6 @@
 import { useCallback } from 'react'
 import { sendAnalyticsEvent, useSession } from '@faststore/sdk'
+import { useLocation } from '@reach/router'
```
igorbrasileiro, line 3: This shouldn't come from gatsby?!

src/sdk/analytics/platform/vtex/index.ts
chrsmutti: Hey @tlgimenes, sorry for the delay. I had a look on our code and what would mean for some of the fields that are missing in this call. At the first moment, `operator` would be the lesser of two evils if it's set to `and` as the default (but it would impact some current and future models we're developing for synonyms).

`misspelled`, is the one that should not be missing and sending them all as `false` could have some impact in the quality of the Top Searches and Autocomplete Suggestions.

So if you could explain why that information that is coming from the API response is not available here, so that we get a clearer picture, it would be good for our internal discussion about this issue.

Thanks for the time! 💪

CHANGELOG.md
tlgimenes: Why do you want me to remove this?

src/sdk/analytics/platform/vtex/search.ts
tlgimenes: it's so much clearer for me 3600

src/sdk/product/useProductLink.ts
tlgimenes: Not really, they don't re-export it.

src/sdk/analytics/platform/vtex/index.ts
chrsmutti: Just had a meeting with the team, and how it's structured today, not having this analytics data sent back to our models will deteriorate the experience compared to other clients that currently send this data to us (other headless accounts, Store Framework, etc). There's no action that we can take at the moment to bypass this.

So if we want the best experience right now, we would need to have this data. We weren't aware of the schema.org limitation. From what I gathered, the response schema is the one used here: [ProductGalleryQuery](https://github.com/vtex-sites/base.store/blob/a9d42e413257b64718de7aab62c3a5614956c791/%40generated/graphql/persisted.json#L4
)?

cc @aprando

tlgimenes: > So if we want the best experience right now, we would need to have this data. We weren't aware of the schema.org limitation. From what I gathered, the response schema is the one used here: [ProductGalleryQuery](https://github.com/vtex-sites/base.store/blob/a9d42e413257b64718de7aab62c3a5614956c791/%40generated/graphql/persisted.json#L4)?

Yes, that's the one. What do you think on releasing this version, and iterating later on? We can try to measure how things goes and if it's really necessary I can try creating an middleware that adds this kind of info for you. 
The middleware could re-do the search and add these values for instance.

aprando: Hi guys!

@tlgimenes unfortunately, as @chrsmutti  commented, if we go up without this information the system will be degraded. It's a fact, we don't even need to deploy a version to measure it.

The main problem is the misspelled attribute. Without it all terms will be considered including misspelled words and even bad words. Are we going to accept the risk of having a misspelled word or having a bad word on top searches or autocomplete?

I also don't think it's a viable option to re-do the query in a middleware. The infrastructure cost for VTEX would be very high. We should use the return of the search performed to get the info. Is the effort for this too big? Maybe it's worth a meeting to discuss and if that's the case our team can help with the implementation.

aprando: Yes! Please don't do it! 🙏

tlgimenes: Ok, so I've removed this bit of the code, let's start with the clicks events first

icazevedo (approved):
LGTM

PR #390, opened 2022-03-15
[FSSS-214] Chore: ignoreOrder warnings from MiniCssExtractPlugin

eduardoformiga:

Solve **some warnings on localhost** from the `mini-css-extract-plugin`. 

## How does it work?

As explained previously on #244 

- Locally, using develop mode (`yarn develop`), the `CSS`s from all pages and components on **localhost** are being extracted into a single CSS file (commons.css) and imported using `<link>` in the `<head>`.

Screen Shot 2022-01-17 at 13 27 24

Thus, as a result of having all CSS together, we are getting **some warnings on localhost** from the `mini-css-extract-plugin`. 

- On the other hand, in production (`yarn build`), as we enabled the CSS code-splitting, we have only the CSS that is being used in the pages/components. Thus optimizing performance and avoiding class collision and CSS order warnings.
In addition, the CSS from the page being used is minified, inside a `<style>` tag in the `<head>`. Thus, the browser avoids one more request to fetch a CSS file.

So, we just have this problem locally then, and, while investigating how to deal with it, we started to test the use of CSS modules to analyze gains regarding the CSS order warnings and performance. [But the result was almost the same as before](https://vtex-dev.atlassian.net/browse/FSSS-127). 

Finally, reading some issues and threads ([#1](https://stackoverflow.com/questions/51971857/mini-css-extract-plugin-warning-in-chunk-chunkname-mini-css-extract-plugin-con/67579319#67579319), [#2](https://stackoverflow.com/questions/63124432/how-do-i-configure-mini-css-extract-plugin-in-gatsby/63128321#63128321), [#3](https://github.com/webpack-contrib/mini-css-extract-plugin#ignoreOrder), [#4](https://github.com/gatsbyjs/gatsby/discussions/30169#discussioncomment-877458), [#5](https://github.com/gatsbyjs/gatsby/issues/11072)) we decide to ignore the order from the `mini-css-extract-plugin` once we have our CSS scoped already and it looks like this is the recommended way to avoid this problem so far.

## How to test it?

Check the console tab on Devtools after running `yarn develop` and double-check that we don't have warnings regarding the CSS order as we had previously. 
You can also check if the warnings were gone on the same console that you run the `yarn develop` command.

## References
https://vtex-dev.atlassian.net/browse/FSSS-214
https://vtex-dev.atlassian.net/browse/FSSS-127

## Checklist

- [x] CHANGELOG entry added


filipewl (approved):
It looks good to me based on what we had discussed!

ArthurTriis1 (approved):

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Scroll lock when transitioning pages on mobile via `SlideOver` component navigation
 - Filter Button specificity on desktop
 - Filter facets are not being selected on mobile
+- Aims to ignore order warnings from `MiniCssExtractPlugin` on develop environment.
```
lariciamota, line 70:
```suggestion
- Ignore order warnings from `MiniCssExtractPlugin` on develop environment.
```

lariciamota (approved):

igorbrasileiro (approved):

eduardoformiga:
Close due to #434

PR #364, opened 2022-02-23
feat: Search history storage

israelswf:
## What's the purpose of this pull request?

This PR aims to save the search history in the client (indexedDB), using the hook useStorage. 

## How does it work?
We've created the `useSearchHistory` hook that returns:
`searchHistory`: returns the content stored in indexedDB
`addToSearchHistory`: can be used to add a new term to the indexedDB
`clearSearchHistory`: clean the indexedDB

Also `MAX_HISTORY_SIZE` sets the maximum terms saved. 

history

## How to test it?

You can do a search and see on indexdb the array of searchs

## References
Jira Task: https://vtex-dev.atlassian.net/browse/FSSS-192

## Checklist

- [ ] CHANGELOG entry added

PR #378, opened 2022-03-08
chore: Add icons to unique svg file and remove from final bundle

igorbrasileiro:
First of all, I want to thank @tlgimenes for the contribution and help.
## What's the purpose of this pull request?
Remove inline SVGs icons from the final bundle and create a unique SVG file with all SVGs as symbols. More information is in the next section.

## How does it work?
All SVGs icons from phosphor react were inlined on the final bundle because they were server-side. To remove those icons inlined, this PR adds a unique SVG file with all icons as symbols, and reuses them with `<use>` tag. Also, this strategy allow style the SVGs with CSS and inheritance. 

Performance results for all pages:
- 🎉  Remove unused javascript from phosphor-react library.

Performance result at home page:
- Decreased the some performance metrics TBT, LCP, TTI on home (check the lighthouse)
- Decreased the `onInitialClientRender` mark in ~17%

Before bundle:
image

After bundle:
image

For more info
https://base.vtex.app/bundle-analyser.html
https://preview-378--base.preview.vtex.app/bundle-analyser.html

https://pagespeed.web.dev/report?url=https%3A%2F%2Fpreview-378--base.preview.vtex.app%2F

## How to test it?
Open the preview, navigate, with fast 3g and without throtling, hover elements.
Check if the element colors, size, widths still the same.

## References
After this approach https://github.com/vtex-sites/base.store/pull/358 i'm trying this one.
https://cloudfour.com/thinks/svg-icon-stress-test/#image-element-with-data-uri

## Checklist

<em>You may erase this after checking them all ;)</em>

- [X] CHANGELOG entry added

PR #355, opened 2022-02-21
feat: Top search component

hellofanny:
## What's the purpose of this pull request?

Adds `Top Search` component, an element that will be part of the `Suggestion` [dropdown component](https://vtex-dev.atlassian.net/browse/FSSS-202).
This component attempts to list the ranking of the top searched items. Since this task  doesn't consider the integration part yet, this component is receiving an array of `searchedItems: LinkItem[]`. It needs to be revised later during the integration.

```
type LinkItem = {
  href: string       //link to redirect when the item is clicked
  name: string    //label to display on the list
}
```

## How does it work?

https://user-images.githubusercontent.com/3356699/155030110-b89b946e-b14f-45b5-bfab-0dcb19eaf930.mov

## How to test it?
There is no preview for it yet. (Because it depends on the Suggestion component)
You can call this component on any page to test it visually.

e.g.:

- Example of `searchedItems`
```
const searchedItems = [
  { name: 'Office Supplies', href: '/' },
  { name: 'Headphones', href: '/' },
  { name: 'Notebooks', href: '/' },
]
```
 
- Calling the component:
`<SuggestionsTopSearch searchedItems={searchedItems} />`

## References

[FSSS-194 - Top Search Component](https://vtex-dev.atlassian.net/jira/software/projects/FSSS/boards/632?selectedIssue=FSSS-194)

## Checklist

<em>You may erase this after checking them all ;)</em>

- [x] CHANGELOG entry added

PR #391, opened 2022-03-15
Create history component

saranicoly:

This PR aims to show the user's search history in base store, which is made by creating a "SearchHistory" component.

How it should look like:
image
How it is actually looking like:
image

## How does it work?

- When the user clicks on "search bar", if there are any previous searches saved in the local storage, **SearchHistory** must appear:

https://user-images.githubusercontent.com/22377378/158483614-c94b66f9-bb40-4742-990b-5e8aef86bc20.mov

- If the user has never searched anything, or if they press the "clear" button, there won't be any search saved in the local storage. In this case, the shown component will be "SuggestionsTopSearch":

https://user-images.githubusercontent.com/22377378/158485506-ba5845a1-e01a-4e1c-b308-501812f2a81e.mov

- When clicking any of the listed searches, the component must do another search with the same name:

https://user-images.githubusercontent.com/22377378/158487623-2bcc4905-0d7e-44a8-ba96-0bc442871c72.mov

## How to test it?

Since the component will be part of the Suggestion's dropdown component, we don't have a specific place to render it yet. That said, you can test SearchHistory by following the steps below:

**Supposing that you want to see it on Navbar:**

- Open the _Navbar.tsx_ file
- Add the following imports:
```
import SearchHistory from 'src/components/search/History/SearchHistory'
import useSearchHistory from 'src/sdk/search/useSeachHistory'
import { SuggestionsTopSearch } from 'src/components/search/Suggestions'
```
- Add `const { searchHistory, clearSearchHistory } = useSearchHistory()` inside the _Navbar()_ function:
image

- Add the following lines to the _SearchInput_ call:

```
onBlur={() => setSearchExpanded(false)}
onFocus={() => setSearchExpanded(true)}
```
image

- Before closing the _section_ tag, add the following lines:

```
{searchExpanded && searchHistory.length > 0 && (
    <SearchHistory onClear={() => clearSearchHistory()} />
 )}
{searchExpanded && searchHistory.length === 0 && (
    <SuggestionsTopSearch
        searchedItems={[
            {
                  href: 'http://localhost:8000/tasty-granite-towels-tasty-18643698/p',
                  name: 'Tasty Granite Towels Tasty',
            },
        ]}
    />
)}
```
image

- After that, you'll be ready to test the component!

### Possible issues

- Some of the functionalities aren't working well on Desktop. This is probably because the component bar is closed before the function can work when we click it. 
- Since the actual rendering of this component will be done in another task and these issues are likely related to the way SearchHistory is being called in the **Navbar**, it was accorded within the team that the investigation would be done later.
- If you face any issue related to this, try to visualize it as a mobile device.

## References

- [Task description](https://vtex-dev.atlassian.net/jira/software/projects/FSSS/boards/632?selectedIssue=FSSS-193)
- [Search history storage](https://github.com/vtex-sites/base.store/pull/364)
- [Search Suggestion Dropdown task](https://vtex-dev.atlassian.net/jira/software/projects/FSSS/boards/632?selectedIssue=FSSS-202)

## Checklist

- [x] CHANGELOG entry added


israelswf (approved):
Great job, Sara ✨

src/components/search/History/SearchHistory.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import { Icon as UIIcon, List as UIList } from '@faststore/ui'
+import Button from 'src/components/ui/Button'
+import './SearchHistory.scss'
+import Link from 'src/components/ui/Link'
+import Icon from 'src/components/ui/Icon'
+import useSearchHistory from 'src/sdk/search/useSeachHistory'
+import { formatSearchState, initSearchState } from '@faststore/sdk'
+
+const doSearch = (term: string) => {
```
filipewl, line 10: Noticed we have this same function on the `SearchInput` component so I wonder we can how we can extract it to a single place.

https://github.com/vtex-sites/base.store/blob/157ac94f2ef63cba213bb78ffb4712145d5d4dec/src/components/common/SearchInput/SearchInput.tsx#L24-L30

src/components/search/History/SearchHistory.tsx
```diff
@@ -0,0 +1,59 @@
...
+        </Button>
+      </div>
+      <UIList variant="ordered">
+        {searchHistory.map((item, index) => (
+          <li key={index}>
+            <Link variant="display" to={doSearch(item)}>
+              <div>
+                <UIIcon
+                  component={<Icon name="Clock" width={18} height={18} />}
+                />
```
filipewl, line 43: Is this the same as the following?
```suggestion
                <UIIcon name="Clock" width={18} height={18} />
```

Same comment for the `ArrowUpRight` below.

If it is, then maybe we can use this shorter version.

src/components/search/History/SearchHistory.scss
```diff
@@ -0,0 +1,37 @@
+@import "src/styles/scaffold";
+
```
filipewl, line 2: This component's structure and styles are very similar to the [Top Search component](https://github.com/vtex-sites/base.store/pull/355/files#diff-7705543ff64105143829013d8db01ea8497c88b73b86b9050f8b1bc1e3a52aafR2). At some point we can try to unify them to reduce some duplication.

src/components/search/History/SearchHistory.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <p className="history__title">History</p>
+        <Button variant="tertiary" onClick={onClear}>
+          Clear
+        </Button>
+      </div>
+      <UIList variant="ordered">
+        {searchHistory.map((item, index) => (
+          <li key={index}>
+            <Link variant="display" to={doSearch(item)}>
+              <div>
```
filipewl, line 40: I was going to comment that block elements such as `div`s were not allowed inside inline elements such as `a`s, but it appears [this rule doesn't apply anymore with HTML 5](https://stackoverflow.com/a/1828032/7970105)! _Today I learned_.

filipewl (commented):
Left some initial comments for us to discuss!

src/components/search/History/SearchHistory.scss
```diff
@@ -0,0 +1,37 @@
...
+    [data-store-link] {
+      display: flex;
+      justify-content: space-between;
+      color: var(--color-neutral-6);
+    }
+
+    [data-store-icon] {
+      margin-right: var(--space-1);
+    }
```
eduardoformiga, line 35: We need to align the rows. Like the first row in the image below:
Screen Shot 2022-03-16 at 19 59 49

So, we need to tweak it a little.
(also considering that the `margin-right` just need to apply on the first icon. If we need to apply to both icons, just ignore the `&:first-child`).
```suggestion
   [data-store-link] {
      display: flex;
      align-items: center;
      justify-content: space-between;
      color: var(--color-neutral-6);

      div {
        display: flex;
        align-items: center;
      }
    }

    [data-store-icon] {
      display: flex;

      &:first-child { margin-right: var(--space-1); }
    }
```

src/components/search/History/SearchHistory.scss
```diff
@@ -0,0 +1,37 @@
...
+  font-size: var(--text-size-3);
+  line-height: 1.5;
+}
+
+.history__header {
+  display: flex;
+  justify-content: space-between;
+}
+
+.history__section {
```
eduardoformiga, line 14: Do we need this class since we have the `data-store-search-history already`?

src/components/search/History/SearchHistory.tsx
eduardoformiga: Nice catch!
I think when we have all search-related components integrated we are going to have a more clear way about how to reuse the same function.

src/components/search/History/SearchHistory.tsx
eduardoformiga: TIL! 😄

eduardoformiga (changes requested):
Good Job! 👏 

I left some comments/suggestions regarding the row alignment.

src/components/search/History/SearchHistory.tsx
saranicoly: I agree with @eduardoformiga. How do you guys feel about adding a task to do that in the next sprint?

src/components/search/History/SearchHistory.scss
saranicoly: Agreed.

src/components/search/History/SearchHistory.tsx
saranicoly: It's not. The icons are being called like that because of these PR's:

-  [SVG unification](https://github.com/vtex-sites/base.store/pull/378)
- [Renaming IconSVG to Icon](https://github.com/vtex-sites/base.store/pull/386)

src/components/search/History/SearchHistory.scss
```diff
@@ -0,0 +1,37 @@
...
+.history__title {
+  align-self: center;
+  font-size: var(--text-size-3);
+  line-height: 1.5;
+}
+
+.history__header {
+  display: flex;
+  justify-content: space-between;
+}
```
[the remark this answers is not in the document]
saranicoly: Good catch. Done!

src/components/search/History/SearchHistory.scss
saranicoly: I think you're right. I'll remove it.

src/components/search/History/SearchHistory.scss
```diff
@@ -0,0 +1,41 @@
+@import "src/styles/scaffold";
+
+.history__header {
```
sergiofontes, line 3: We should target `data-attributes` instead of `classes`.

src/components/search/History/SearchHistory.scss
```diff
@@ -0,0 +1,37 @@
...
+  display: flex;
+  justify-content: space-between;
+}
+
+.history__section {
+  padding: var(--space-1) var(--space-3) var(--space-2);
+  background-color: var(--bg-neutral-lightest);
+}
+
+[data-store-search-history] {
```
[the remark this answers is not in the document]
sergiofontes: Yeah, absolutely. Let's aim for the lesser style specificity as possible.

src/components/search/History/SearchHistory.scss
```diff
@@ -0,0 +1,37 @@
...
+  ol {
```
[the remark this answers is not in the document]
sergiofontes: If they're ordered in a way that resembles the history timeline, then it's `ol`.

sergiofontes (approved):

src/components/search/History/SearchHistory.tsx
eduardoformiga: Agree with you!

eduardoformiga (approved):

lucasfp13 (approved):
Great job, LGTM! 💯 

Just a small thing I noticed is that the CSS file name is not following the pattern we have in the project. Instead using `SearchHistory.scss` you can leave as `search-history.scss` to keep consistency.

src/components/search/History/SearchHistory.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <div data-store-search-history-header>
+        <h4 data-store-search-history-title>History</h4>
+        <Button variant="tertiary" onClick={onClear}>
+          Clear
+        </Button>
+      </div>
+      <UIList variant="ordered">
+        {searchHistory.map((item, index) => (
+          <li data-store-search-history-item key={index}>
+            <Link variant="display" to={doSearch(item)}>
```
filipewl, line 39: I forgot what we concluded on this, but I remember mentioning that the arrow-up-right icon is usually an indicator of the page opening in a new tab. If we're going this way, does it make sense to add the following?

```suggestion
            <Link variant="display" to={doSearch(item)} target="_blank">
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target

filipewl (commented):
LGTM! Just one more question about opening the searches in a new tab.

src/components/search/History/SearchHistory.tsx
saranicoly: Sure!

PR #396, opened 2022-03-17
feat: Adds new Badge interactive variation

lucasfp13:

This PR intends to add a new `Badge` variant called `interactive`. This will be useful when the user needs an action to close/remove badges.

## How does it work?

**Default interactive badge:**

Screen Shot 2022-03-17 at 16 48 04Screen Shot 2022-03-17 at 16 48 13

**Small interactive badge:**
Screen Shot 2022-03-17 at 16 46 34Screen Shot 2022-03-17 at 16 48 26

## References

[FSSS 204 - Creates Badge Interactive variation](https://vtex-dev.atlassian.net/browse/FSSS-204?atlOrigin=eyJpIjoiZDgzY2IzZmU1ODY5NGY1YjlkNTljODlhOTQ5ZGZlMTAiLCJwIjoiaiJ9)

## Checklist

- [x] CHANGELOG entry added

### DIFF
--- a/src/components/ui/Badge/Badge.tsx
+++ b/src/components/ui/Badge/Badge.tsx
@@ -1,18 +1,27 @@
 import { Badge as UIBadge } from '@faststore/ui'
 import type { ReactNode } from 'react'
 import React from 'react'
+import IconButton from 'src/components/ui/IconButton'
+import Icon from 'src/components/ui/Icon'
 
 import './badge.scss'
 
-export type BadgeVariants = 'info' | 'highlighted' | 'neutral'
+export type BadgeVariants = 'info' | 'highlighted' | 'neutral' | 'interactive'
 
 type Props = {
   small?: boolean
   variant?: BadgeVariants
   children: ReactNode
+  onClose?: () => void
 }
 
-const Badge = ({ small = false, variant, children, ...otherProps }: Props) => {
+const Badge = ({
+  small = false,
+  variant,
+  children,
+  onClose,
+  ...otherProps
+}: Props) => {
   return (
     <UIBadge
       className="badge"
@@ -20,7 +29,21 @@ const Badge = ({ small = false, variant, children, ...otherProps }: Props) => {
       data-store-badge-variant={variant}
       {...otherProps}
     >
-      {children}
+      <span>{children}</span>
+      {variant === 'interactive' && (
+        <IconButton
+          onClick={onClose}
+          aria-label="Remove badge"
+          icon={
+            <Icon
+              name="X"
+              weight="bold"
+              width={small ? 12 : 16}
+              height={small ? 12 : 16}
+            />
+          }
+        />
+      )}
     </UIBadge>
   )
 }
--- a/src/components/ui/Badge/Badge.tsx
+++ b/src/components/ui/Badge/Badge.tsx
@@ -6,20 +6,22 @@ import Icon from 'src/components/ui/Icon'
 
 import './badge.scss'
 
-export type BadgeVariants = 'info' | 'highlighted' | 'neutral' | 'interactive'
+export type BadgeVariants = 'info' | 'highlighted' | 'neutral'
 
 type Props = {
   small?: boolean
   variant?: BadgeVariants
   children: ReactNode
   onClose?: () => void
+  interactive?: boolean
 }
 
 const Badge = ({
-  small = false,
   variant,
   children,
   onClose,
+  small = false,
+  interactive = false,
   ...otherProps
 }: Props) => {
   return (
@@ -27,10 +29,11 @@ const Badge = ({
       className="badge"
       data-store-badge={small ? 'small' : ''}
       data-store-badge-variant={variant}
+      data-store-badge-interactive={interactive}
       {...otherProps}
     >
       <span>{children}</span>
-      {variant === 'interactive' && (
+      {(interactive || onClose) && (
         <IconButton
           onClick={onClose}
           aria-label="Remove badge"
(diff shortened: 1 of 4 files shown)


src/components/ui/Badge/Badge.tsx
```diff
@@ -1,26 +1,49 @@
 import { Badge as UIBadge } from '@faststore/ui'
 import type { ReactNode } from 'react'
 import React from 'react'
+import IconButton from 'src/components/ui/IconButton'
+import Icon from 'src/components/ui/Icon'
 
 import './badge.scss'
 
-export type BadgeVariants = 'info' | 'highlighted' | 'neutral'
+export type BadgeVariants = 'info' | 'highlighted' | 'neutral' | 'interactive'
```
eduardoformiga, line 9: Hmm, I think the `interactive Badge` shouldn't be a new variant. Since the `Badge` can be `info`/`highlighted`/`neutral` and `interactive` at the same time. Maybe we can deal with the `interactive` checking the `onClose` callback existence so that we can deal with both _states_? What do you think?

lucasfp13: Yeah, you are right. I miss the point of using the action together with the other cases

filipewl: Agreed!

I was thinking about how we could apply [the needed different styles when in interactive mode](https://github.com/vtex-sites/base.store/pull/396/files#diff-a1eda55d5a5721215d3eed0d9797b4df0781a3c9c53d51b9c7c76f5b5521f980R39) assuming `data-store-badge-variant=interactive` won't be possible. I think we can either:

1. Put the interactive information in another data attribute (e.g. `data-store-badge-mode=interactive`).
2. Join the variant **and** interactive information in the same data attribute (e.g. `data-store-badge-mode="info interactive"`) and match the interactive in CSS via the `[attr*=value]` selector (e.g. `data-store-badge-variant*=interactive`).

> <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#:~:text=followed)%20by%20value.-,%5Battr*%3Dvalue%5D,-Represents%20elements%20with">[attr*=value]</a>
> Represents elements with an attribute name of attr whose value contains at least one occurrence of value within the string.

src/components/ui/Badge/badge.scss
```diff
@@ -27,6 +36,24 @@
...
+
+  [data-store-icon-button] {
+    width: var(--tap-min-size);
+    height: auto;
+    color: var(--color-neutral-7);
+    border-top-right-radius: var(--border-radius-pill);
+    border-bottom-right-radius: var(--border-radius-pill);
+
+    &:hover {
+      background-color: var(--color-secondary-1);
```
filipewl, line 52: Will this be the same color for differently colored badges? I'm assuming it's always a darker tone than the variant's background color. But then we'd have to specify the different colors for each variant I guess. Maybe there's a way to keep it DRY using a color function, but not sure yet.

heitorado (approved):
I agree with @eduardoformiga [review and suggestion](https://github.com/vtex-sites/base.store/pull/396/files#r829480186) - Apart from that, everything looks good 👍🏼

filipewl (approved):
Looking smooth!!! Left some points for discussion.

src/components/ui/Badge/Badge.tsx
```diff
@@ -1,26 +1,49 @@
...
   return (
     <UIBadge
       className="badge"
       data-store-badge={small ? 'small' : ''}
       data-store-badge-variant={variant}
       {...otherProps}
     >
-      {children}
+      <span>{children}</span>
+      {variant === 'interactive' && (
```
ArthurTriis1, line 33: Following the @eduardoformiga comment, we can show interactive icon just if the onClose event is pass

```suggestion
      {onClose && (
```

src/components/ui/Badge/Badge.tsx
```diff
@@ -10,17 +12,41 @@ type Props = {
...
     <UIBadge
       className="badge"
       data-store-badge={small ? 'small' : ''}
       data-store-badge-variant={variant}
+      data-store-badge-interactive={interactive}
       {...otherProps}
     >
-      {children}
+      <span>{children}</span>
+      {(interactive || onClose) && (
```
ArthurTriis1, line 36: This way it generates a little inconsistency in the style, because if we add onClose and not add interactive we would have the button, but not the style, like in the gif:


badge

ArthurTriis1: I think we have two options, show the button only when "interactive" is "pass" or add the data-attribute when onClose is pass too

src/components/ui/Badge/Badge.tsx
```diff
@@ -10,17 +12,41 @@ type Props = {
   small?: boolean
   variant?: BadgeVariants
   children: ReactNode
+  onClose?: () => void
+  interactive?: boolean
 }
```
ArthurTriis1, line 17: Also, we can try this:

-  if have interactive, onClose is not required
- if havent interactive, onClose is not permitted
- if have onClose, interactive is required

```suggestion
  small?: boolean
  variant?: BadgeVariants
  children: ReactNode
  onClose?: () => void
  interactive?: boolean
} & BadgeOnClose

type BadgeOnClose =
  | {
      interactive: true
      onClose?: () => void
    }
  | {
      interactive?: false
      onClose?: never
    }
```

ArthurTriis1 (approved):

src/components/ui/Badge/badge.scss
lucasfp13: nice catch @filipewl, after you raised this point I talked to Flip and his advice is to use black as the hover color with low opacity, I'll apply that

filipewl (approved):

src/components/ui/Badge/Badge.tsx
hellofanny: I agreed with Filipe's suggestion. It seems a nice approach for this scenario.

PR #408, opened 2022-03-22
chore: Remove inline styles

tlgimenes:
This PR replaces inline <style/> tags by <link/> tags.

## How does it work?
Inlining styles is good technique for decreasing FCP. It, however, increases TBT, since the browser processes both CSS and HTML in one go, instead of processing them on two different tasks.
This PR moves inline styles to an isolated stylesheet and removes CSS code splitting for the sake of simplicity. 

FCP shouldn't be penalized that much since this CSS is hashed and has an `immutable` cache header, which is heavily cached by the CDN.

Moving inline styles to a custom sheet is not natively supported by gatsby, so a hack had to be made. This hack is described in here: https://github.com/gatsbyjs/gatsby/issues/1526

## How to test it?
Nothing should have changed visually. However, PSI score should be bigger than master

## Checklist
- [x] CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `SearchHistory` component.
 
 ### Changed
-
+- Move inline styles to external stylesheet to improve TBT
 - Changed ProductGallery and EmptyGallery styles to make the search results page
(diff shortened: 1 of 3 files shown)


igorbrasileiro:
Did it increased the CLS? Or still the same?

igorbrasileiro (approved):

PR #415, opened 2022-03-23
chore: Supersede stylelint-config-rational-order

heitorado:

This PR replaces `stylelint-config-rational-order`. The plugin has become problematic with several bugs, reported on the npm audit, and in [this issue](https://github.com/carbon-design-system/carbon/issues/8450). It is also not being updated for a long time.

The replacement for it is `stylelint-config-recess-order` ([link](https://github.com/stormwarning/stylelint-config-recess-order))

## How does it work?
It removes the `yarn` entry for the previous package and adds the new one, applying the required configurations.
~After running `stylelint --fix` across all stylesheets, many of them had the order of some attributes changed. For those reviewing with more experience in CSS, it is necessary to confirm if the new order is correct, following the `rational-order` that we aim to use in this codebase.~

**UPDATE:** As requested by @sergiofontes the stylesheets will not be updated in this PR, to avoid conflicts with an upcoming change that will affect style files.

## How to test it?
Nothing should have changed visually in the store.
Check if the order of CSS rules is still correct after the change.

## Checklist
- [x] CHANGELOG entry added


sergiofontes (changes requested):
Could you please remove the CSS files sorting, @heitorado? We'll be releasing a PR that will have an impact on the entire CSS base. So, to avoid unnecessary (for the time being) conflicts, we could defer this reordering task to another PR. Could you do it?

heitorado:
Sure thing.

sergiofontes (approved):

filipewl (approved):

eduardoformiga (approved):
LGTM! 
I need to get used to the new order tweaks, though! hahah 😄

igorbrasileiro (approved):

PR #420, opened 2022-03-24
Fix:  slide over scroll behavior

eduardoformiga:

This PR aims to fix the scroll behavior while opening some SlideOver related components (Filter, Menu, Minicart) on the mobile version. It also refactors some code so that we can prioritize the use of optional chaining instead of if-clauses.

## How does it work?

1. The first step was to identify what was affecting the scroll behavior: `SlideOver` component while adding the `no-scroll` `class` to the `layout` `div`.
2. Looking into the CSS applied: 
```css
#layout.no-scroll {
  position: fixed;
  top: 0;
  right: 0;
  left: 0;
  overflow: hidden;
}
```
The `position: fixed` makes the `layout div` scroll to the top. So first of all we need to remove this piece of code and keep just the `overflow: hidden;`.

3. But if we remove the `fixed position` on the `layout div`, we are going to have another problem: the scroll from the `body` element while the filter is opened. We need to keep the `body` frozen instead of the `layout div`. So we need to adapt this code to deal with the `body` instead of the `layout div`.

|BEFORE|AFTER|
|-|-|
|BefaXkYYL8659|AfOuXNS5Q80E|

## How to test it?

Double-check in the preview link if we have the scroll behavior while opening some `SlideOver` related components (Filter, Menu, Minicart) on the mobile version.

## References

https://vtex-dev.atlassian.net/browse/FSSS-212
https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Operators/Optional_chaining
https://medium.com/react-camp/how-to-fight-the-body-scroll-2b00267b37ac
https://futurefoundry.co/blog/disable-body-scroll-ios/

## Checklist

- [x] CHANGELOG entry added


filipewl (approved):
Awesome! Tested on https://sfj-cbe68cb--base.preview.vtex.app/ (on my Android as well) and it's not scrolling to the top anymore.

src/styles/global/layout.scss
```diff
@@ -50,11 +50,7 @@ main,
 }
 
 // Avoids double scroll issue on the page when overlay is open
-#layout.no-scroll {
-  position: fixed;
-  top: 0;
-  right: 0;
-  left: 0;
+.no-scroll {
   overflow: hidden;
```
hellofanny, line 54: This doesn't work very well on iOS safari :(
I remember trying a few approaches to this problem before.
Have you tried something with `touch-action:none`?

https://user-images.githubusercontent.com/3356699/160032184-fa95c50d-b384-4420-b41b-b1f1a00c738e.mov

eduardoformiga: Nice catch! I'll look into new approaches to make it works on iOS Safari.

hellofanny: It is still not working [properly](https://sfj-e43e031--base.preview.vtex.app/) 😞 Let's use another approach for that.

hellofanny: I tried a similar [approach](https://github.com/vtex-sites/base.store/pull/443 ) with position fixed. It[ locks the scroll inside the modal,](https://preview-443--base.preview.vtex.app/) but we always get the scroll position back to the top. Should we try to solve it using some external library? 🤔

hellofanny: I tried a similar [approach](https://github.com/vtex-sites/base.store/pull/443 ) with position fixed. It[ locks the scroll inside the modal,](https://preview-443--base.preview.vtex.app/) but we always get the scroll position back to the top. Should we try to solve it using some external library? 🤔

igorbrasileiro (changes requested):
Add tests to prevent this bug in future.

src/styles/global/layout.scss
eduardoformiga: well, the solution above solves the bug when touching the `overlay`. But the bug remains while scrolling into the `minicart` just using iOS Safari. I also have tried different approaches to that problem but without success.

As this is a very specific known problem only on iOS Safari. I think we can go ahead. What do you think?

eduardoformiga:
> 

Nice! Done [here](https://github.com/vtex-sites/base.store/pull/420/commits/7effa6b00bdff84d518c7b411886b0a6863f0a90).

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - CSS Warnings (#434)
 - Fix alert banner colors (#442)
+- Fix `SlideOver` scroll background behavior (#420)
```
filipewl, line 51: Minor suggestion so [the last version comes first](https://keepachangelog.com/en/1.0.0/#:~:text=The%20latest%20version%20comes%20first).
```suggestion
- Fix `SlideOver` scroll background behavior (#420)
- CSS Warnings (#434)
- Fix alert banner colors (#442)
```

filipewl (approved):
Thanks for all the research you did to try to fix this.
🙇

renatamottam (approved):
Awesome! 👏 👏 👏

PR #422, opened 2022-03-25
chore: Simplify filters

tlgimenes:
This PR uses the `useReducer` hook to simplify the state management and hopefully reduce complexity on the Filters component.

## How does it work?
Using multiple `useState` on the same component can be troublesome and difficult to reason. `useReducer` is usually a better fit for this case. Also, I've notice that a [small feature](https://github.com/vtex/faststore/pull/1189) on `@faststore/sdk` could help us reduce complexity a lot. 

## How to test it?
Make sure no regression was made, and filters behave like they were before.

## Checklist
- [x] CHANGELOG entry added


igorbrasileiro:
What do you think about using a Map<string, Facet>, where the string is the facet key?

tlgimenes:
> What do you think about using a `Map<string, Facet & { expanded?: boolean }>`, where the string is the facet key?

Instead of the array? I'm not really sure because there may be multiple facets with the same key. The key+value is unique, not only the key

tlgimenes:
> > > What do you think about using a `Map<string, Facet & { expanded?: boolean }>`, where the string is the facet key?
> > 
> > 
> > Instead of the array? I'm not really sure because there may be multiple facets with the same key. The key+value is unique, not only the key
> 
> Could merge the value from the same key, couldn't?! It wouldn't affect the behavior, afaik.

Not sure. Imagine the case where two brands are selected. You would have the same key `brand` with two different values, `brand-1` and `brand-2`

src/components/search/Filter/Filter.tsx
```diff
@@ -32,154 +30,34 @@ interface FilterProps {
...
-          selectedFacets={selectedFacets}
-          filteredFacets={filteredFacets}
-          indicesExpanded={indicesExpanded}
+          indicesExpanded={expanded}
           onFacetChange={toggleFacet}
-          onAccordionChange={onAccordionChange}
-          onAccordionItemMount={onAccordionItemMount}
+          onAccordionChange={(index) =>
+            dispatch({ type: 'toggleExpanded', payload: index })
+          }
```
igorbrasileiro, line 60: Make sense to wrap it as a callback function to prevent re-render the Facets component?

src/components/search/Filter/Filter.tsx
```diff
@@ -202,34 +80,41 @@ function Filter({
...
+                  type: 'selectFacets',
+                  payload: selectedFacets,
+                })
+
                 dismissTransition.current?.()
               }}
             />
           </header>
           <Facets
+            facets={facets}
```
igorbrasileiro, line 93: There's a way to reuse the same Facets component instead of rendering two?

src/components/search/Filter/useFilter.ts
```diff
@@ -0,0 +1,118 @@
...
+  const {
+    state: { selectedFacets },
+  } = useSearch()
+
+  const [{ selected, expanded }, dispatch] = useReducer(reducer, null, () => ({
+    expanded: new Set([]),
+    selected: selectedFacets,
+  }))
+
+  const selectedMap = useMemo(
```
igorbrasileiro, line 80: Using a HasMap isn't better than an object?

src/components/search/Filter/Filter.tsx
tlgimenes: Not really, because the callbacks are different

src/components/search/Filter/Filter.tsx
tlgimenes: I don't think it makes any difference on this case, but I can do it

src/components/search/Filter/useFilter.ts
tlgimenes: yes, I'll change it

src/components/search/Filter/Filter.tsx
tlgimenes: Yeah, I tried doing this now but had some issues, let's leave it like this

src/components/search/Filter/useFilter.ts
tlgimenes: done

igorbrasileiro:
Results for 20 running PSI tests
Page result - https://base.vtex.app/office/
| Metric | Mean | Standard deviation |
|--------|--------|--------|
| Cumulative Layout shift (CLS) | 0.00 | 0.00 |
| First Contentful Paint (FCP) | 1550.04 | 4412.94 |
| First Contentful Paint 3g (FCP) | 3087.54 | 17261.20 |
| Largest Contentful Paint (LCP) | 1795.24 | 129925.89 |
| Time to Interactive (TTI) | 3696.01 | 89215.45 |
| Total Blocking Time (TBT) | 617.46 | 32647.79 |
| Performance score | 0.825 | 0.001875 |
| JavaScript Execution Time | 1960.61 | 71026.08 |
| Speed Index | 1975.97 | 25792.81 |

Page result - https://sfj-5d6a78d--base.preview.vtex.app/office/
| Metric | Mean | Standard deviation |
|--------|--------|--------|
| Cumulative Layout shift (CLS) | 0.00 | 0.00 |
| First Contentful Paint (FCP) | 1301.65 | 18723.70 |
| First Contentful Paint 3g (FCP) | 2681.92 | 169908.37 |
| Largest Contentful Paint (LCP) | 1598.06 | 16541.32 |
| Time to Interactive (TTI) | 3538.96 | 83616.60 |
| Total Blocking Time (TBT) | 549.18 | 19384.91 👀  |
| Performance score | 0.854 | 0.001344 👀  |
| JavaScript Execution Time | 1800.07 | 41073.98 |
| Speed Index | 1588.78 | 3976.43 |

igorbrasileiro (approved):
Just remember to bump @faststore/api to a release version.

PR #426, opened 2022-03-26
feat: Creates Regionalization modal

lucasfp13:

This PR intends to create the `RegionalizationModal`, but with some caveats:
- the `PostalCodeInput` is not ready yet, so right now we are using the current not styled version;
- the `Don't know my Postal Code` link needs to point to the right URL, so we need to decide which one we will use;

With this component, users will be able to set/inform their current location to be able to see available products for their region.

## How does it work?

**NOTE: The `RegionalizationModal` is being rendered in the `Navbar` just for testing purposes, so when reviewing that we do not need to worry about it because it will be removed as soon as these changes are approved.**

When opened, the `RegionalizationModal` has a fade-in effect and looks like this:

- Desktop version
Screen Shot 2022-03-25 at 21 03 16

- Mobile version
Screen Shot 2022-03-25 at 21 03 33

A quick demonstration of how motion effect looks like: 

https://user-images.githubusercontent.com/15722605/161068860-79894afc-1f3c-4e88-9358-f1c2c9604b5f.mov

## References

[FSSS 230 - Create regionalization Modal](https://vtex-dev.atlassian.net/browse/FSSS-230)

## Checklist]

- [x] CHANGELOG entry added


src/components/ui/SlideOver/SlideOver.tsx
```diff
@@ -18,74 +18,34 @@ interface SlideOverProps extends HTMLAttributes<HTMLDivElement> {
...
-        handleClose()
+        onModalClose()
       }}
       data-slide-over
       data-slide-over-direction={direction}
       data-slide-over-size={size}
-      data-slide-over-state={fadeType}
-      onTransitionEnd={handleTransitionEnd}
+      data-slide-over-state={fade}
+      onTransitionEnd={() => fade === 'out' && onDismiss?.()}
```
tlgimenes, line 48: is it possible to run into a race condition where `fade` is `in` and the `onDismiss` is not called, leaving the modal open?

src/sdk/ui/modal/Provider.tsx
```diff
@@ -0,0 +1,56 @@
+import React, {
+  useMemo,
+  useState,
+  useContext,
+  useCallback,
+  createContext,
+} from 'react'
+import type { PropsWithChildren } from 'react'
+
+import './provider.scss'
```
hellofanny, line 10: hmm, I think it's better to identify with the prefix otherwise sounds too generic, how about renaming to `modal-provider.scss`?

hellofanny, line 10: hmm, I think it's better to identify with the prefix otherwise sounds too generic, how about renaming to `modal-provider.scss`?

src/sdk/ui/modal/Provider.tsx
```diff
@@ -0,0 +1,56 @@
...
+      fade,
+      onModalOpen,
+      onModalClose,
+    }),
+    [fade, onModalOpen, onModalClose]
+  )
+
+  return (
+    <ModalContext.Provider value={value}>
+      <div data-scroll={fade === 'out'}>{children}</div>
```
hellofanny, line 41: Is this data attr added to avoid the double scroll on the page?

hellofanny, line 41: Is this data attr added to avoid the double scroll on the page?

src/components/regionalization/RegionalizationModal/regionalization-modal.scss
```diff
@@ -0,0 +1,100 @@
+@import "src/styles/scaffold";
+
+[data-modal-overlay] {
```
hellofanny, line 3: We are not scoping these styles in the component, so in this case, this will be duplicated. I believe we can remove it and keep the different background style.

image

hellofanny, line 3: We are not scoping these styles in the component, so in this case, this will be duplicated. I believe we can remove it and keep the different background style.

image

src/components/regionalization/RegionalizationModal/regionalization-modal.scss
```diff
@@ -0,0 +1,100 @@
...
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 3;
```
hellofanny, line 9:
```suggestion
```

hellofanny, line 9:
```suggestion
```

hellofanny: It's such a slight difference 🤔 Maybe we could double-check with the design team this background opacity.

hellofanny: It's such a slight difference 🤔 Maybe we could double-check with the design team this background opacity.

src/components/regionalization/RegionalizationModal/regionalization-modal.scss
```diff
@@ -0,0 +1,100 @@
...
+  background-color: rgb(0 0 0 / 25%);
```
hellofanny, line 10: It's such a slight difference 🤔 Maybe we could double-check with the design team this background opacity.

hellofanny, line 10: It's such a slight difference 🤔 Maybe we could double-check with the design team this background opacity.

src/components/regionalization/RegionalizationModal/regionalization-modal.scss
```diff
@@ -0,0 +1,100 @@
...
+
+    [data-regionalization-modal-description] {
+      color: var(--fs-color-text-light);
+    }
+
+    .regionalization-modal__button {
+      position: absolute;
+      top: 0;
+      right: 0;
+      background-color: transparent;
```
hellofanny, line 51: Can we just keep the default background? Otherwise we will lost the focus effect/indicator.

```suggestion
```

hellofanny, line 51: Can we just keep the default background? Otherwise we will lost the focus effect/indicator.

```suggestion
```

hellofanny: |Changed| Default|
|-|-| 
no-bg|default

hellofanny: |Changed| Default|
|-|-| 
no-bg|default

hellofanny:
This modal looks great 🤩!!! I really like the idea of using a provider to replace the previous callbacks and handling the in/out behavior; looks neater than before 🙈 
I haven't finished reviewing yet, but I've left some notes. I'll come back to review more deeply, okay? 🤓

src/components/ui/SlideOver/SlideOver.tsx
lucasfp13: I believe it is not, we are safe against the race condition by only calling `onDismiss` after transition/animation ends and, before that, checking for the right `fade` value

src/components/regionalization/RegionalizationModal/regionalization-modal.scss
```diff
@@ -0,0 +1,93 @@
...
+    width: calc(100vw / 3);
+    min-width: calc(var(--fs-grid-breakpoint-desktop) / 3);
+  }
+
+  &[data-regionalization-modal-state="in"] {
+    animation: motion-in .2s ease-in;
+  }
+
+  &[data-regionalization-modal-state="out"] {
+    animation: motion-out .2s ease-out;
```
filipewl, line 27: Some equivalent tokens I found:
```suggestion
    animation: motion-in var(--fs-transition-timing) ease-in;
  }

  &[data-regionalization-modal-state="out"] {
    animation: motion-out var(--fs-transition-timing) ease-out;
```

filipewl (approved):
Nice refactoring! Everything else looks good to me.
👏

src/styles/global/layout.scss
```diff
@@ -49,15 +49,6 @@ main,
   padding-left: var(--fs-grid-padding);
 }
 
-// Avoids double scroll issue on the page when overlay is open
-#layout.no-scroll {
```
eduardoformiga, line 53: Nicee! I'm touching this piece of code in another PR but I can sync this change there 👀

eduardoformiga (approved):
Good Job! 
The code looks more clean and readable now ✨

PR #431, opened 2022-03-30
feat: Move shelf client side

tlgimenes:
This PR moves `<ProductShelf/>` and `<ProductTiles/>` to the client side

## How does it work?
Shelves and tiles are usually used to display a certain product collection. The product availability/price on this collection depends on the channel the user is in. Since the product list to be displayed depends on a browser defined property (channel), we need to make these components dynamic by rendering them on the client.

This PR uses the `useProductsQuery` to fetch products on the browser. 

The big visual difference is that, instead of loading the shelf, it loads the loading state of the shelf, for them querying and displaying the right product. A before/after comparison can be seen below:

<div style="display: flex">
before
after  
</div>

## How to test it?
Make sure all shelves are ok.

## Checklist
- [x] CHANGELOG entry added


src/components/sections/ProductShelf/ProductShelf.tsx
```diff
@@ -1,34 +1,40 @@
...
+  }
+
   return (
     <Section
       className={`layout__section ${withDivisor ? 'shelf__divisor' : ''}`}
     >
       <h2 className="text__title-section layout__content">{title}</h2>
       <div data-product-shelf-content>
-        <ProductShelfSkeleton loading={products.length === 0}>
+        <ProductShelfSkeleton loading={products == null}>
```
igorbrasileiro, line 33:
```suggestion
        <ProductShelfSkeleton loading={products === null}>
```

src/components/sections/ProductShelf/ProductShelf.tsx
```diff
@@ -1,34 +1,40 @@
...
           <ul data-product-shelf-items className="layout__content">
-            {products.map((product, idx) => (
-              <li key={`${product.id}`}>
-                <ProductCard product={product} index={idx + 1} />
+            {products?.edges.map((product, idx) => (
```
igorbrasileiro, line 35:
```suggestion
            {products.edges.map((product, idx) => (
```
Not needed, since it's already verified. If the ts complain, need to fix the first conditional at line 23.

src/components/sections/ProductTiles/ProductTiles.tsx
```diff
@@ -28,21 +28,27 @@ const getRatio = (products: number, idx: number) => {
...
   return (
     <Section className="layout__section layout__content">
       <h2 className="text__title-section">{title}</h2>
       <div>
-        <ProductTilesSkeleton variant="wide" loading={products.length === 0}>
+        <ProductTilesSkeleton variant="wide" loading={!products}>
           <Tiles>
-            {products.map((product, idx) => (
-              <Tile key={product.id}>
+            {products?.edges.map((product, idx) => (
```
igorbrasileiro, line 44: The same comment i left before.

src/components/sections/ProductShelf/ProductShelf.tsx
tlgimenes: not really, we need to load on undefined too

igorbrasileiro: Use or operator, make the code more readable

src/components/sections/ProductShelf/ProductShelf.tsx
igorbrasileiro: Did you fixed the line 23?

tlgimenes: what do you think we should use for the conditional of line 23? We can only return `null` when there is no products, a.k.a, products.length === 0. When it's loading, we should render the component for the skeleton take place

igorbrasileiro: Ok, np, but this is weird form, is it for you?

igorbrasileiro (approved):

src/components/sections/ProductShelf/ProductShelf.tsx
igorbrasileiro: Oh i see...
products?.edge.length can be undefined, right? You need to check like this
```tsx
!products || products.edge.length === 0
```

tlgimenes: yes, but that's the thing. I don't want it to return null when products is undefined. When it's undefined it needs to render the skeleton

PR #432, opened 2022-03-30
feat: Style the search dropdown

filipewl:

Style the search dropdown according to [this Figma specification](https://www.figma.com/file/xPRr75q5Qm9SlZ963hBh8O/Handoff-v3-(Search)?node-id=2%3A58234):

**Mobile**

If there is History|If there isn't History
-|-
image|image

**Desktop**
If there is History|If there isn't History
-|-
image|image

## How does it work?

Previews for the implementation:

Mobile|Desktop
-|-
CleanShot 2022-03-30 at 09 57 23|CleanShot 2022-03-30 at 09 59 44

## What's missing?

- [ ] The dropdown's backdrop for the mobile.
- [ ] Show the `Suggestions` component when typing.
- [ ] The logic when to show the history or top search.

## How to test it?

Visit [this preview link](https://sfj-ae2f2aa--base.preview.vtex.app/) and interact with the search input.

## References

- [Jira FSSS-202 Search Suggestion Dropdown](https://vtex-dev.atlassian.net/browse/FSSS-202)


filipewl (commented):
Some helper comments.

israelswf:
Just made a test with a long text, works good. 👍 
Just noticed the icon vertically aligned with the middle of the text (not sure if it is something to be handled in this PR, though)
image

israelswf:
Also noticed when we click some item in the history, the `searchInput` keeps the old content. I guess it would be interesting if we passed the clicked term to the input (or just clean the input, maybe).

https://user-images.githubusercontent.com/10616243/160937127-5498a164-ac67-4ebe-8f7a-58850f753f95.mp4

renatamottam (changes requested):
Great work, Wands! We're almost there :)

A couple improvements:
1. Left and right paddings should be removed from both `[data-store-search-history-item] [data-store-link]` and `data-store-search-history-header`.
2. `data-store-search-history-item` hover should be the same as the `.suggestions__item`
3. `.suggestions__item [data-store-link]` should fill the whole `width` so the person can select an option without having to click on the name.

|Padding|Hover|
|-|-|
|image|image|

hellofanny:
> Great work, Wands! We're almost there :)
> 
> A couple improvements:
> 
> 1. Left and right paddings should be removed from both `[data-store-search-history-item] [data-store-link]` and `data-store-search-history-header`.
> 2. `data-store-search-history-item` hover should be the same as the `.suggestions__item`
> 3. `.suggestions__item [data-store-link]` should fill the whole `width` so the person can select an option without having to click on the name.
> 
> Padding	Hover
> image	image

Same thing for the header / title. Back then, my idea was to share these common styles/classes between those components (`Suggestions`, `SuggestionsTopSearch` and `SearchHistory`. I still think we should do so. However, we need to rename these components' prefixes to make sense.

filipewl:
> Also noticed when we click some item in the history, the `searchInput` keeps the old content. I guess it would be interesting if we passed the clicked term to the input (or just clean the input, maybe).
> 
>  20220330184733391.mp4

@israelswf, you're right! I addressed it [here](https://github.com/vtex-sites/base.store/pull/432/commits/bfdbe0de388467951f59ddc8628a647ed519aa19) so when a dropdown item is selected, the search input is replaced by its term.

CleanShot 2022-03-30 at 21 22 57

PR #434, opened 2022-03-30
fix: CSS Warnings

tlgimenes:
This PR is part in a series of PRs aiming to improve Developer Perception about base.store. This PR focusses on removing those annoying CSS Warnings and solves possible issues with import order and class crashes.

## How does it work?
The warnings were being thrown because the way we were importing CSS files. Importing all css files at once in a single `components.scss` file solves the issue. Also, this makes it work on next.js and other frameworks.

If you are adding a new component, please import the component's css into the `components.scss` instead on the component itself. 

If you are wondering if performance will be harmed in any ways, the answer is no, because we were already using a single CSS anyways.

Other solutions using CSS modules may be attempted in the future for better css splitting if it's needed down the line

## How to test it?
Make sure all styles appear on the store. Also, make sure the warnings went away.

## Checklist
- [x] CHANGELOG entry added


igorbrasileiro (commented):
WDYT about adding a disclaimer about it at README.md?

eduardoformiga (commented):
Niceee! I think this is the best thinking in the long term. 👏 

@sergiofontes @renatamottam do you think this can affect the new Theme structure?

tlgimenes:
> WDYT about adding a disclaimer about it at README.md?

Nice catch! I just changed the section about adding styles to a component.

eduardoformiga:
close #390

eduardoformiga:
Closes #390

gatsby-browser.js
```diff
@@ -1,10 +1,9 @@
+import './src/styles/fonts.css'
+import './src/styles/global/layout.scss'
 import './src/styles/global/resets.scss'
-
 import './src/styles/global/tokens.scss'
```
sergiofontes, line 4: `tokens.scss` must come before all other CSS files (excepting `fonts.scss`).
Regarding the `global`s, I suggest the following order:
1. `tokens.scss`
2. `resets.scss`
3. `typography.scss`
4. `layout.scss`

gatsby-browser.js
```diff
@@ -1,10 +1,9 @@
+import './src/styles/fonts.css'
```
sergiofontes, line 1: Wouldn't be more performant to load these `fonts` within the `<head>`? As we currently make use of Google Fonts, Base Store could benefit from Google's built-in optimizations, such as:
* [Combination of requests](https://www.smashingmagazine.com/2019/06/optimizing-google-fonts-performance/#3-combine-requests);
* [DNS Preconnect](https://www.smashingmagazine.com/2019/06/optimizing-google-fonts-performance/#preconnect).

sergiofontes (changes requested):
I'm all for it if it helps to improve the overall developer experience. But I'm curious how we're going to improve the CSS splitting by using CSS Modules, given that this isn't their primary purpose. I'm afraid of these bundled stylesheets because they obstruct rendering and lower overall LH scores, but we can use this solution while we investigate other options.

gatsby-browser.js
tlgimenes: stringifying this file content into the <head> you really improve performance! nice catch! However, we are not using google fonts now, we are using local ones, right?

tlgimenes:
> I'm all for it if it helps to improve the overall developer experience. But I'm curious how we're going to improve the CSS splitting by using CSS Modules, given that this isn't their primary purpose. I'm afraid of these bundled stylesheets because they obstruct rendering and lower overall LH scores, but we can use this solution while we investigate other options.

Yes, you are right! CSS Modules and code splitting are two different things. On this PR, I don't change any code splitting whatsoever. If you pay attention closely, the generated stylesheets before/after this PR are the same. The only difference is that we generate this sheet without any warnings.
A future PR can be made to split the CSS.

sergiofontes:
> Yes, you are right! CSS Modules and code splitting are two different things. On this PR, I don't change any code splitting whatsoever. If you pay attention closely, the generated stylesheets before/after this PR are the same. The only difference is that we generate this sheet without any warnings.
> A future PR can be made to split the CSS.
Got it, thanks!

sergiofontes (approved):

gatsby-browser.js
sergiofontes: You're right: serving fonts in the same server is better than connecting to Google, right? Thanks!

igorbrasileiro (approved):

gatsby-browser.js
igorbrasileiro: Afaik, defining the fonts families and the usage of these new fonts, increase performance, decreasing the recalculate styles that decrease the tbt, and can decrease the CLS if exists.

PR #392, opened 2022-03-16
feat: Add session query and run session query after change postal code

igorbrasileiro:
## What's the purpose of this pull request?
This PRs adds the session region for VTEX platform and runs the session query when the postal code changes.

## How does it work?
When the user changes the postal code, fetches a new channel that contains regionId as value.

## How to test it?
Set a valid postal code in the postal code and press `Enter`

Also, test the search, add to cart.

## Checklist
- [x] CHANGELOG entry added

PR #450, opened 2022-04-04
Updates CHANGELOG to match base.store's March Release Notes

danzanzini:

This PR updates the CHANGELOG so it matches it's [March Release Notes](https://faststore.dev/releases/2022/04/01/basestore). 
I've also added the identifier of the PR to each CHANGELOG entry so we don't need to rely on git blame to identify the PR responsible for the entry.

## How does it work?

Each entry that was merged before #392 was added to the 0.2.0 section. A new [Github release](https://github.com/vtex-sites/base.store/releases/tag/v0.2.0) was also added to reflect this state.

Each entry after this is still in the Unreleased section.

## How to test it?

Check if the CHANGELOG makes sense and doesn't have any wrong information.

## References
https://keepachangelog.com/en/1.0.0/

## Checklist

- [x] CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,94 +8,113 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- Add preloadQuery function
-- Add hideUnavailableItems at store.config
-- Sections component with `content-visibility: auto`
-- Webpack Bundle analyzer
-- `GatsbyLink` to `Link` ui component.
-- `Skeleton` loading components.
-- `SuggestionsTopSearch` component
-- `PostalCodeInput` component and `usePostalCode` hook.
-- `SuggestionProductCard` component.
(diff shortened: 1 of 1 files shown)


Mariana-Caetano (approved):
Awesome @danzanzini! 🎉 Just one thing to add is that the PRs from February were mentioned in the [Full Changelog section of the Release Notes](https://github.com/vtex-sites/base.store/releases) (at the beginning of it) since the main focus was the March PRs. Either way LGTM.

danzanzini:
@Mariana-Caetano Do you think it's worth creating a section below the 0.2.0 for entries from February?

Mariana-Caetano:
> @Mariana-Caetano Do you think it's worth creating a section below the 0.2.0 for entries from February?

@danzanzini It sounds like a good idea, but do you mean also creating a 0.1.2 release to it?

danzanzini:
A proper release is not necessary but I can create a tag to reflect the state of the code

PR #484, opened 2022-04-18
chore: Remove the Lighthouse check via GH Action

filipewl:

The "Wait for build to succeed" step is constantly failing because our WebOps' "Build" check takes some time to start and finish, so the GH Actions spends at least 10 minutes waiting before failing, requiring a manual re-run when it's known then "Build" check has finished. These minutes of waiting are quickly consuming our GH plan quota, sometimes forcing us to upgrade to the next plan.

<small>_[Example](https://github.com/vtex-sites/base.store/runs/6029510252?check_suite_focus=true) of the GH Action waiting for the Build check to finish before proceeding_</small>
CleanShot 2022-04-18 at 15 21 50

## How does it work?

By removing the config file for the GH Action.

## How to test it?

This PR itself must not run the GH Action.

## References

See [this internal Slack thread](https://vtex.slack.com/archives/C01U97YRFFG/p1649853593310519).


igorbrasileiro (approved):

danzanzini:
I agree with removing it. It generates a lot of noise. 

Since we already have [lighthouse running at vtex-sites](https://github.com/vtex-sites/base.store/pull/484/checks?check_run_id=6079180018), do you think it's worth working into other alternatives?

filipewl:
> I agree with removing it. It generates a lot of noise.
> 
> Since we already have [lighthouse running at vtex-sites](https://github.com/vtex-sites/base.store/pull/484/checks?check_run_id=6079180018), do you think it's worth looking at other alternatives?

@danzanzini, I don't think it's needed to look for alternatives since we're already running it via WebOps. 

This GH Action was originally added because we were having instability issues with the WebOps one in the past, but now it has been stabilized. I think we're good to remove it now!

eduardoformiga (approved):

danzanzini (approved):
Nice. Do we have to disable the action in Github to get rid of this pending check, right?

filipewl:
> Nice. Do we have to disable the action in Github to get rid of this pending check, right?

@danzanzini, I already disabled the workflow [here](https://github.com/vtex-sites/base.store/actions/workflows/main.yml), but it's still expecting to run and it's blocking the merge.
😠 

CleanShot 2022-04-19 at 11 57 02

saranicoly (approved):

PR #401, opened 2022-03-21
chore: Improve images performance

tlgimenes:
## What's the purpose of this pull request?
This PR drops `gatsby-plugin-image` in favor of custom image component.
`gatsby-plugin-image` is great for creating responsive images when using a source plugin or when handling local images. Since we don't use local images and we will stop using source plugins with Gatsby v4, `gatsby-plugin-image` doesn't really help us. 
This PR removes this `gatsby-plugin-image` and uses `thumbor` server directly. A nice side effect is that base.store gets more framework independent in the process.

## How does it work?
This PR adds an `Image` component alongside an `useImage` hook. This hook generates all necessary tags for a responsive `` tag. The `<Image/>` component uses this hook with `preload` capabilities for LCP images. 

`useImage` hook requires an explicit `src`, `width` and `height` for the target image. After setting the desired `width` and `height`, the hook will generate 3 breakpoints, with `[1, 2, 3] * width`. You can help browser decide which image to use by using this `sizes` attribute. More info [in here](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images)

`<Image/>` component uses the `useImage` hook alongside preload tags for improving the LCP image. To prelaod an image, just add a `preload` attribute with the Image component: 
```
<Image src="https:...." width={200} height={200} preload />
```

> Make sure there is only one preloaded image per page and this image matches the page's LCP.

In the future, we will also do a `Picture` component that allows rendering images with different aspect ratios depending on the user device (mobile/desktop). This is necessary for creating a banner component.

## How to test it?
Everything should be the same

## References
For more info on responsive images:
- https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

## Checklist
- [x] CHANGELOG entry added

PR #489, opened 2022-04-20
Feat: Creates Picture Component

eduardoformiga:

Creates the `Picture` component and `useSources` hook.

## How does it work?

While analyzing some of our client's projects, I noticed a problem regarding the way the images were loading when we have multiples breakpoints, like mobile and desktop.

Taking into consideration that we may want to load different images according to the device (Art Direction) and we were controlling what will be shown using CSS (`display-mobile` and `hidden-mobile` classes), the browser was downloading at least two images, one time for mobile and another for desktop.

Example:
```js
<Banner>
  <Image
    className="display-mobile"
    alt={altMobile}
    src={srcMobile}
    width={widthMobile}
    height={heightMobile}
  />
  <Image
    className="hidden-mobile"
    alt={altDesktop}
    src={srcDesktop}
    width={widthDesktop}
    height={heightDesktop}
  />
</Banner>
```

Since the HTML5 provides the `<picture>` tag along with `<source>`, `srcset`, and `media` queries, we can pass some of them and the browser will download just the best option for the device screen width and pixel density. So, we can pass these options instead of controlling everything using CSS, what decreases the DOM size, CSS classes, and also we can save the bandwidth downloading smaller images. Thus, speeds up page load times.

`<picture>` will download all image sources as needed and dynamically show the appropriate at the time, even if the user changes the dimensions of the page.  Its main use is Art Direction (when you want to show a differently shaped picture under different conditions. Maybe you want to show a cropped version of the picture on mobile or scale some text (e.g. small text that would become unreadable when scaled).

Why not use the `Image` component?

The `Image` component is recommended when we just have one image (mobile or desktop) and we would like to have **multiples copies of the same image**, preserving the aspect ratio for all image copies with more or fewer pixels. When we would like to have **different images for different screens**, the `<picture>` tag is recommended.

Multiples copies of the same image:
Screen Shot 2022-04-20 at 16 59 33

Different images for different screens:
Screen Shot 2022-04-20 at 16 59 18
Screen Shot 2022-04-20 at 17 03 36

## How to test it?

So far on base.store we don't have the Art Direction problem (maybe on Hero or Tiles afterward?). If we decided to change the implementation of these components we can test better.

Example of use:
```js
<Picture
  sources={[
    {
      src: srcMobile,
      width: widthMobile,
      height: heightMobile,
      media: '(max-width: 767px)',
    },
    {
      src: srcDesktop,
      width: widthDesktop,
      height: heightdesktop,
      media: '(min-width: 768px)',
    },
  ]}
  img={{
    src: srcFallback,
    width: widthFallback,
    height: heightFallback,
    alt: altFallback,
    loading: "lazy",
  }}
/>
```

Final markup example:
```js
<picture>
  <source srcset="https://assets.vtex.app/328x435/image328.png 328w, https://assets.vtex.app/656x870/image656.png 656w, https://assets.vtex.app/984x1305/image984.png 984w" media="(max-width: 767px)">
  <source srcset="https://assets.vtex.app/1304x636/image1304.png 1304w, https://assets.vtex.app/2608x1272/image2608.png 2608w, https://assets.vtex.app/3912x1908/image3912.png 3912w" media="(min-width: 768px)">
  <img src="https://store.vtexassets.com/assets/image.png" width="1304" height="636" alt="Alternative text" loading="lazy">
</picture>
```

## References

https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_image
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source
https://blog.bitsrc.io/why-you-should-use-picture-tag-instead-of-img-tag-b9841e86bf8b
https://www.w3schools.com/html/html_images_picture.asp
https://web.dev/preload-responsive-images/#preload-and-lesspicturegreater
https://github.com/vtex-sites/base.store/pull/401

## Checklist

- [x] CHANGELOG entry added

### DIFF
--- a/src/components/ui/Image/Picture.tsx
+++ b/src/components/ui/Image/Picture.tsx
@@ -46,5 +46,4 @@ function Picture({
   )
 }
 
-Picture.displayName = 'Picture'
 export default memo(Picture)
--- a/src/components/ui/Image/Picture.tsx
+++ b/src/components/ui/Image/Picture.tsx
@@ -1,6 +1,6 @@
-import type { HTMLAttributes } from 'react'
-import { memo } from 'react'
+import React, { memo } from 'react'
 import { Helmet } from 'react-helmet'
+import type { HTMLAttributes } from 'react'
 
 import { useSources } from './useSources'
 import type { SourceOptions } from './useSources'
@@ -23,7 +23,7 @@ function Picture({
     <picture>
       {sources.map(({ srcSet, media, preloadLinks }, index) => {
         return (
-          <>
+          <React.Fragment key={index}>
             {preload && (
               <Helmet
                 link={preloadLinks.map(
@@ -38,7 +38,7 @@ function Picture({
               />
             )}
             <source key={index} srcSet={srcSet} media={media} />
-          </>
+          </React.Fragment>
         )
       })}
       <img {...imgProps} alt={imgProps.alt} />
--- a/src/components/ui/Image/Picture.tsx
+++ b/src/components/ui/Image/Picture.tsx
@@ -1,0 +1,27 @@
+import type { HTMLAttributes } from 'react'
+import { memo } from 'react'
+
+import { useSources } from './useSources'
+import type { SourceOptions } from './useSources'
+import type { ImageOptions } from './useImage'
+
+interface Props extends HTMLAttributes<HTMLPictureElement> {
+  sources: SourceOptions[]
+  img: ImageOptions
+}
+
+function Picture({ sources: sourceOptions, img: imgProps }: Props) {
+  const sources = useSources(sourceOptions)
+
+  return (
+    <picture>
+      {sources.map(({ srcSet, media }, index) => {
+        return <source key={index} srcSet={srcSet} media={media} />
+      })}
+      <img {...imgProps} alt={imgProps.alt} />
+    </picture>
+  )
+}
+
+Picture.displayName = 'Picture'
+export default memo(Picture)
--- a/src/components/ui/Image/Picture.tsx
+++ b/src/components/ui/Image/Picture.tsx
@@ -1,5 +1,6 @@
 import type { HTMLAttributes } from 'react'
 import { memo } from 'react'
+import { Helmet } from 'react-helmet'
 
 import { useSources } from './useSources'
 import type { SourceOptions } from './useSources'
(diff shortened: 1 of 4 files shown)


src/components/ui/Image/Picture.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <picture>
+      {sources.map(({ srcSet, media }, index) => {
+        return <source key={index} srcSet={srcSet} media={media} />
+      })}
+      <img {...imgProps} alt={imgProps.alt} />
+    </picture>
+  )
+}
+
+Picture.displayName = 'Picture'
```
filipewl, line 26: Do we need this for debugging purposes?

src/components/ui/Image/useSources.ts
```diff
@@ -0,0 +1,38 @@
...
+  width: number
+  height: number
+  media: string
+}
+
+export interface SourceOptions extends SourceProps {
+  options?: ThumborOptions
+}
+
+const FACTORS = [1, 2, 3]
```
filipewl, line 18: Are these multiplying factors always these numbers?

eduardoformiga: These FACTORS were based on the previously implemented `useImage`. 

 These are the most common values based on the device's density. You can type `devicePixelRatio` on `devtools` console to see the value for each selected device.

src/components/ui/Image/Picture.tsx
eduardoformiga: I don't think so. Removed [here](https://github.com/vtex-sites/base.store/pull/489/commits/73f66cd686f5d1ec5379c6048cf6100405131457).

src/components/ui/Image/Picture.tsx
```diff
@@ -0,0 +1,49 @@
...
+                  (href: string) =>
+                    ({
+                      as: 'image',
+                      rel: 'preload',
+                      href,
+                      media,
+                    } as any)
+                )}
+              />
+            )}
```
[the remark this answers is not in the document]
eduardoformiga: I follow [this approach](https://web.dev/preload-responsive-images/#preload-and-lesspicturegreater), it looks like there is no tailored solution for the `<picture>` though.

src/components/ui/Image/useSources.ts
```diff
@@ -0,0 +1,45 @@
...
+  extends SourceHTMLAttributes<HTMLSourceElement> {
+  preloadLinks: string[]
+}
+
+const FACTORS = [1, 2, 3]
+
+export const useSources = (
+  sourcesOptions: SourceOptions[]
+): SourceHTMLAttributesPreload[] => {
+  return useMemo(() => {
```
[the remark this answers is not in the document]
eduardoformiga: That's because we can deal with more dimensions than just desktop and mobile. 
And we are dealing with the preload links as well.

